PR #9, opened 2020-11-03
Use new BackendMemory API

deadeyegoodwin:


src/onnxruntime.cc
```diff
@@ -1324,8 +1325,10 @@ ModelInstanceState::SetStringInputTensor(
   BackendMemory* input_memory;
   RESPOND_ALL_AND_RETURN_IF_ERROR(
       responses, request_count,
-      BackendMemory::CreateWithFallback(
-          model_state_->TritonMemoryManager(), TRITONSERVER_MEMORY_CPU_PINNED,
+      BackendMemory::Create(
+          model_state_->TritonMemoryManager(),
+          {BackendMemory::AllocationType::CPU_PINNED_POOL,
+           BackendMemory::AllocationType::CPU},
```
CoderHam, line 1331: I think the order should be CPU_PINNED_POOL, CPU_PINNED, CPU

CoderHam (changes requested):

src/onnxruntime.cc
deadeyegoodwin: You usually don't want to allocate pinned memory on the fly as it is very slow. So if you can't get it from the pool it is likely better to just use non-pinned memory and let the cuda driver sort it out.

CoderHam: Right. Makes sense.

CoderHam (approved):

PR #16, opened 2020-11-23
Fix autofill. Only skip if both inputs and outputs are specified

GuanLuo:


src/onnxruntime.cc
```diff
@@ -436,7 +440,7 @@ ModelState::AutoCompleteMaxBatch(
   // Set max-batch-size to 1 if we have determined that batching is
```
deadeyegoodwin, line 440: Need to add a test for case where batching is not supported but the user specified a max_batch_size > 1. In that case should return an appropriate error.

deadeyegoodwin (changes requested):

deadeyegoodwin (approved):

PR #36, opened 2021-03-29
Add log warning to notify user of max_batch_size autofill policy

CoderHam:


src/onnxruntime.cc
```diff
@@ -447,6 +447,12 @@ ModelState::AutoCompleteMaxBatch(
       ModelConfig().Find("max_batch_size", &mbs_value);
       mbs_value.SetInt(1);
       SetMaxBatchSize(1);
+      LOG_MESSAGE(
+          TRITONSERVER_LOG_INFO,
+          (std::string("autofilled max_batch_size to 1 for model '") + Name() +
+           "' since batching is supporrted but no max_batch_size is specified "
+           "in model configuration")
```
GuanLuo, line 454: Maybe changing it to warning and add that the user should specify max_batch_size if the user wants to utilize auto completion but with a larger max batch size.

CoderHam: Yes that is a good idea

GuanLuo (approved):

PR #37, opened 2021-03-30
Improve error messages to be more clear

CoderHam:


src/onnxruntime.cc
```diff
@@ -795,11 +793,9 @@ ModelInstanceState::ValidateTypedSequenceControl(
...
           TRITONSERVER_ERROR_INVALID_ARG,
           (std::string("unable to load model '") + model_state_->Name() +
-           "', sequence control '" + tensor_name +
-           "', the model expects data-type " +
-           OnnxDataTypeName(iit->second.type_) +
-           " but the model configuration specifies data-type " +
-           tensor_datatype)
+           "', configuration expects datatype " + tensor_datatype +
+           " for sequence control '" + tensor_name + "', model provides " +
+           OnnxDataTypeName(iit->second.type_))
```
GuanLuo, line 798: The original error message for this and above already mentioned the config / model difference

deadeyegoodwin: Agreed, I would keep the above 2 the same as they were

CoderHam: Sure, make this change

PR #41, opened 2021-04-26
Expand ONNXRuntime backend for Jetson/zero-container build

CoderHam:


tanmayv25 (dismissed):

CMakeLists.txt
```diff
@@ -364,38 +371,40 @@ install(
...
-
-install(
-  DIRECTORY
-    ${CMAKE_CURRENT_BINARY_DIR}/onnxruntime/bin/
-  USE_SOURCE_PERMISSIONS
-  DESTINATION ${CMAKE_INSTALL_PREFIX}/backends/onnxruntime
-)
-
-if (NOT WIN32)
+if(TRITON_ONNXRUNTIME_DOCKER_BUILD)
```
deadeyegoodwin, line 374: You shouldn't disable install for non-docker build.

deadeyegoodwin (changes requested):

CMakeLists.txt
CoderHam: This install copies over the core onnxruntime libraries and not the backend itself. For Jetson we already built these in advance. 
I will add a comment that clarifies the same.

deadeyegoodwin (approved):

Issue #45, opened 2021-06-01
Memory leak in ONNX runtime backend


Tabrizian:
**Description**
Memory leak in the ONNX runtime backend.

**Triton Information**
What version of Triton are you using?
main branch

Are you using the Triton container or did you build it yourself?
container

**To Reproduce**
The `L0_infer_valgrind` test can be used for repro. When running the test, `TEST_VALGRIND=1` should be set to 1.

**Expected behavior**

 There should be no memory leaks.
 ```
 ==18067== 20,736 bytes in 1,296 blocks are definitely lost in loss record 184,459 of 184,803
==18067==    at 0x483E0F0: memalign (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==18067==    by 0x483E212: posix_memalign (in /usr/lib/x86_64-linux-gnu/valgrind/vgpreload_memcheck-amd64-linux.so)
==18067==    by 0x14EA987A4: onnxruntime::utils::DefaultAlloc(unsigned long) (in /opt/tritonserver/backends/onnxruntime/libonnxruntime.so.1.7.1)
==18067==    by 0x14E096CDC: OrtApis::GetBoundOutputValues(OrtIoBinding const*, OrtAllocator*, OrtValue***, unsigned long*) (in /opt/tritonserver/backends/onnxruntime/libonnxruntime.so.1.7.1)
==18067==    by 0x14DD88D6A: triton::backend::onnxruntime::ModelInstanceState::ReadOutputTensors(unsigned long, std::vector<char const*, std::allocator<char const*> > const&, TRITONBACKEND_Request**, unsigned int, std::vector<TRITONBACKEND_Response*, std::allocator<TRITONBACKEND_Response*> >*) (in /opt/tritonserver/backends/onnxruntime/libtriton_onnxruntime.so)
==18067==    by 0x14DD8BA27: triton::backend::onnxruntime::ModelInstanceState::ProcessRequests(TRITONBACKEND_Request**, unsigned int) (in /opt/tritonserver/backends/onnxruntime/libtriton_onnxruntime.so)
==18067==    by 0x14DD8DCA2: TRITONBACKEND_ModelInstanceExecute (in /opt/tritonserver/backends/onnxruntime/libtriton_onnxruntime.so)
[...]

PR #46, opened 2021-06-01
Fixing mem leak reported in #45

askhade:
After all the OrtValues returned from GetBoundOutputValues are individually released, the output buffer which is allocated using the specified instance of the allocator must be freed using the same allocator.

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -608,6 +608,7 @@ class ModelInstanceState : public BackendModelInstance {
   // on this instance.
   std::vector<OrtValue*> input_tensors_;
   std::vector<OrtValue*> output_tensors_;
+  OrtValue** output_buffer_;
   std::vector<BackendMemory*> input_tensor_memories_;
 };
 
@@ -726,12 +727,29 @@ ModelInstanceState::ReleaseOrtRunResources()
   }
   input_tensors_.clear();
 
+  // first release the Ortvalues
   for (auto& tensor : output_tensors_) {
     if (tensor != nullptr) {
       ort_api->ReleaseValue(tensor);
     }
   }
   output_tensors_.clear();
+
+  // next release the allocated buffer using the specified allocator
+  if (output_buffer_) {
+    auto free_status =
+        ort_api->AllocatorFree(default_allocator_, output_buffer_);
+    output_buffer_ = nullptr;
+    if (free_status != nullptr) {
+      LOG_MESSAGE(
+          TRITONSERVER_LOG_ERROR,
+          (std::string("onnx runtime allocator free error:") +
+           std::to_string(ort_api->GetErrorCode(free_status)) +
+           ort_api->GetErrorMessage(free_status))
+              .c_str());
+      ort_api->ReleaseStatus(free_status);
+    }
+  }
 
   for (BackendMemory* mem : input_tensor_memories_) {
     delete mem;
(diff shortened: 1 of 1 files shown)


src/onnxruntime.cc
```diff
@@ -1661,11 +1679,10 @@ ModelInstanceState::ReadOutputTensors(
   }
 
   size_t output_count = 0;
-  OrtValue** output_values = nullptr;
   RESPOND_ALL_AND_RETURN_IF_ORT_ERROR(
       responses, request_count,
       ort_api->GetBoundOutputValues(
-          io_binding_, default_allocator_, &output_values, &output_count));
+          io_binding_, default_allocator_, &output_buffer_, &output_count));
```
deadeyegoodwin, line 1685: Aren't you overwriting output_buffer_ here each time? Are you relying on it always returning the same buffer? Is that a valid assumption?

deadeyegoodwin (changes requested):

src/onnxruntime.cc
```diff
@@ -608,6 +608,7 @@ class ModelInstanceState : public BackendModelInstance {
   // on this instance.
   std::vector<OrtValue*> input_tensors_;
   std::vector<OrtValue*> output_tensors_;
+  OrtValue** output_buffer_;
```
deadeyegoodwin, line 611: You should initialize in construtor to nullptr.

deadeyegoodwin (changes requested):

src/onnxruntime.cc
askhade: I dont exactly understand your concern here
GetBoundOutputValues call will allocate the output buffer using the specified instance of the allocator which we then free before exiting ProcessRequest.

deadeyegoodwin: I see, I missed that call to ReleaseOrtRunResources

askhade: The only thing changing in this PR is now we are rightfully freeing the output_buffer_ before exiting ProcessRequests. Earlier we were not and hence the mem leak.

deadeyegoodwin: OK, can you initiliaze output_buffer_ and I will start the CI

pranavsharma:
At some point I suggest we move to using unique_ptrs for all ORT resources as it's very easy to forget to call Release. Even better switch to C++ wrappers.

src/onnxruntime.cc
```diff
@@ -726,13 +727,30 @@ ModelInstanceState::ReleaseOrtRunResources()
...
   for (auto& tensor : output_tensors_) {
     if (tensor != nullptr) {
       ort_api->ReleaseValue(tensor);
     }
   }
   output_tensors_.clear();
 
+  // next release the allocated buffer using the specified allocator
+  if (output_buffer_) {
+    auto free_status =
```
pranavsharma, line 740: can use unique_ptr to avoid calling Release.

deadeyegoodwin (approved):

askhade:
> At some point I suggest we move to using unique_ptrs for all ORT resources as it's very easy to forget to call Release. Even better switch to C++ wrappers.

I was planning to address this (moving to unique_ptrs for ORT resources) in a separate PR. Regarding moving to c++ wrappers I had not considered that yet... Will talk to David before proceeding on this.

PR #49, opened 2021-06-25
Modify backend to be host policy aware

GuanLuo:
See PR on backend repo as reference https://github.com/triton-inference-server/backend/pull/28

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -1209,7 +1209,8 @@ ModelInstanceState::ProcessRequests(
   bool cuda_copy = false;
   BackendInputCollector collector(
       requests, request_count, &responses, model_state_->TritonMemoryManager(),
-      model_state_->EnablePinnedInput(), CudaStream());
+      model_state_->EnablePinnedInput(), CudaStream(),
+      HostPolicyName().c_str());
   SetInputTensors(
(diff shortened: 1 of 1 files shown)


deadeyegoodwin (approved):

askhade:
Can you add an example of host policy... What is the motivation for this change?

GuanLuo:
It is used in the case where multiple copies of the input data already reside in different GPU devices (or memory that is "closer" to particular model instance, i.e. on NUMA configured system). Host policy is an abstraction that helps with getting the input data pointer that is the closest to the model instance. The model instance will be set to a host policy name on creation, by default it is the string representation of the device that the instance is on (i.e. "gpu_0"), but it can be set to any string in config.pbtxt

Currently only applications that use Triton C API may utilize this, where it preloads the data to different GPU devices, and when a request arrives, it looks up the corresponding data pointers on different devices and attach them to the Triton request with corresponding host policies, then the backend will retrieve the data from the same host policy.

Host policy also provides [other types of optimization](https://github.com/triton-inference-server/server/blob/main/docs/optimization.md#numa-optimization) in addition to the one mentioned above, but those are not visible from the backend's perspective.

askhade (approved):

PR #60, opened 2021-08-13
Update ortbackend

askhade:
Enables trt 8.0 for ORT backend. TRT 8.0 support is not yet available in a released version of ort therefore using tensorrt-8.0 branch in ort instead of the rel-* branches until ORT 1.9 release.

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -832,8 +832,8 @@ ModelInstanceState::ValidateBooleanSequenceControl(
   std::string tensor_datatype;
   RETURN_IF_ERROR(GetBooleanSequenceControlProperties(
       sequence_batching, model_state_->Name(), control_kind, required,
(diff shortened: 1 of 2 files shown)


src/onnxruntime.cc
```diff
@@ -832,7 +832,8 @@ ModelInstanceState::ValidateBooleanSequenceControl(
   std::string tensor_datatype;
   RETURN_IF_ERROR(GetBooleanSequenceControlProperties(
       sequence_batching, model_state_->Name(), control_kind, required,
-      &tensor_name, &tensor_datatype, nullptr, nullptr, nullptr, nullptr));
+      &tensor_name, &tensor_datatype, nullptr, nullptr, nullptr, nullptr,
+      nullptr, nullptr));
```
askhade, line 836: will remove this as it should not be part of r21.08... it is only required on main

PR #62, opened 2021-08-19
ort trt update

askhade:
Enables trt 8.0 for ORT backend. TRT 8.0 support is not yet available in a released version of ort therefore using tensorrt-8.0 branch in ort instead of the rel-* branches until ORT 1.9 release.


deadeyegoodwin (approved):

askhade:
I missed removing it for windows. We can remove --use_openmp from windows in master. It is OK to go ahead with 21.08 with openmp enabled for windows

PR #64, opened 2021-09-02
enable more trt options

askhade:
Enabling following options for TRT EP

- ORT_TENSORRT_INT8_ENABLE
- ORT_TENSORRT_INT8_CALIBRATION_TABLE_NAME
- ORT_TENSORRT_INT8_USE_NATIVE_CALIBRATION_TABLE
- ORT_TENSORRT_ENGINE_CACHE_ENABLE
- ORT_TENSORRT_CACHE_PATH

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -268,7 +268,7 @@ ModelState::LoadModel(
                         trt_options.trt_fp16_enable = 1;
                       } else if (value_string == "INT8") {
                         trt_options.trt_int8_enable = 1;
-                        trt_options.trt_int8_calibration_table_name = &int8_calibration_table_name;
+                        trt_options.trt_int8_calibration_table_name = int8_calibration_table_name.c_str();
                       } else if (value_string != "FP32") {
                         RETURN_ERROR_IF_FALSE(
                             false, TRITONSERVER_ERROR_INVALID_ARG,
@@ -286,7 +286,7 @@ ModelState::LoadModel(
                     } else if (param_key == "int8_calibration_table_name") {
                       RETURN_IF_ERROR(params.MemberAsString(
                           param_key.c_str(), &int8_calibration_table_name));
-                      trt_options.trt_int8_calibration_table_name = &int8_calibration_table_name;
+                      trt_options.trt_int8_calibration_table_name = int8_calibration_table_name.c_str();
                     } else if (param_key == "int8_use_native_calibration_table") {
                       RETURN_IF_ERROR(params.MemberAsString(
                           param_key.c_str(), &value_string));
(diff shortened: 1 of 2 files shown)


src/onnxruntime.cc
```diff
@@ -278,6 +283,28 @@ ModelState::LoadModel(
                           value_string, &max_workspace_size_bytes));
                       trt_options.trt_max_workspace_size =
                           static_cast<size_t>(max_workspace_size_bytes);
+                    } else if (param_key == "int8_calibration_table_name") {
+                      RETURN_IF_ERROR(params.MemberAsString(
+                          param_key.c_str(), &int8_calibration_table_name));
+                      trt_options.trt_int8_calibration_table_name = &int8_calibration_table_name;
```
efeguney, line 289:
```suggestion
                      trt_options.trt_int8_calibration_table_name = int8_calibration_table_name.c_str();
```
It looks like trt_int8_calibration_table_name  is cont char*, and assigning string* to char* doesn't seem right.

src/onnxruntime.cc
```diff
@@ -278,6 +283,28 @@ ModelState::LoadModel(
...
+                      RETURN_IF_ERROR(params.MemberAsString(
+                          param_key.c_str(), &value_string));
+                      bool enable_cache;
+                      RETURN_IF_ERROR(ParseBoolValue(
+                          value_string, &enable_cache));
+                      trt_options.trt_engine_cache_enable = enable_cache;
+                    } else if (param_key == "trt_engine_cache_path") {
+                      RETURN_IF_ERROR(params.MemberAsString(
+                          param_key.c_str(), &trt_engine_cache_path));
+                      trt_options.engine_cache_path = &trt_engine_cache_path;
```
efeguney, line 307:
```suggestion
                      trt_options.engine_cache_path = trt_engine_cache_path.c_str();
```
Same as above.

src/onnxruntime.cc
```diff
@@ -264,6 +266,9 @@ ModelState::LoadModel(
                           param_key.c_str(), &value_string));
                       if (value_string == "FP16") {
                         trt_options.trt_fp16_enable = 1;
+                      } else if (value_string == "INT8") {
+                        trt_options.trt_int8_enable = 1;
+                        trt_options.trt_int8_calibration_table_name = &int8_calibration_table_name;
```
efeguney, line 271:
```suggestion
                        trt_options.trt_int8_calibration_table_name = int8_calibration_table_name.c_str();
```
trt_int8_calibration_table_name seems to be const char* based on the definition of OrtTensorRTProviderOptions

efeguney (changes requested):
Thank you for the PR. It looks like we may need to change the assignments to const char* members of OrtTensorRTProviderOptions.

src/onnxruntime.cc
deadeyegoodwin: I'm surprised that the compile didn't fail due to the type mismatch

askhade: I did not pick the changes correctly ... I had staged this file already and then added this fix after seeing compilation error... but forgot to stage this fix before the commit...

deadeyegoodwin (approved):

Issue #69, opened 2021-09-17
ORT backend causes Triton to crash for a failed inference run 


tanmayv25:
**Description**
When `ort_api->RunWithBinding(session_, runOptions_, io_binding_)` returns a non-zero status code, the backend correctly calls SendErrorForResponses to return the error back to the client. However, it does not thwart the further progress of the pipeline and leads to segmentation fault. This causes the entire server to crash which is especially problematic in production environment.

**Triton Information**
21.08 containers

Are you using the Triton container or did you build it yourself?
Triton container

**To Reproduce**

We have a bert mini model which we can not share but the issue may be reproduced by similar model. When running inference with wrong inputs the client does report the encountered error:

![image (1)](https://user-images.githubusercontent.com/10909321/133840344-7963995f-8c7f-45e8-9901-b595797ad238.png)

But the server crashes:

![image (2)](https://user-images.githubusercontent.com/10909321/133840421-cb214ecc-7750-4fd3-a076-880ac1565c43.png)

See the backtrace of the segmentation fault below:

```
Thread 16 "tritonserver" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7f3854b39000 (LWP 510)]
0x00007f3857a361e6 in OrtApis::CastTypeInfoToTensorInfo(OrtTypeInfo const*, OrtTensorTypeAndShapeInfo const**) () from /opt/tritonserver/backends/onnxruntime/libonnxruntime.so.1.8.1
(gdb) bt
#0  0x00007f3857a361e6 in OrtApis::CastTypeInfoToTensorInfo(OrtTypeInfo const*, OrtTensorTypeAndShapeInfo const**) () from /opt/tritonserver/backends/onnxruntime/libonnxruntime.so.1.8.1
#1  0x00007f386819a669 in triton::backend::onnxruntime::ModelInstanceState::ReadOutputTensors(unsigned long, std::vector<char const*, std::allocator<char const*> > const&, TRITONBACKEND_Request**, unsigned int, std::vector<TRITONBACKEND_Response*, std::allocator<TRITONBACKEND_Response*> >*) ()
   from /opt/tritonserver/backends/onnxruntime/libtriton_onnxruntime.so
#2  0x00007f386819cb9a in triton::backend::onnxruntime::ModelInstanceState::ProcessRequests(TRITONBACKEND_Request**, unsigned int) ()
   from /opt/tritonserver/backends/onnxruntime/libtriton_onnxruntime.so
[...]

PR #71, opened 2021-09-21
Return early if ortrun bumps into a inference error

jcwchen:
**Description**
Make OrtRun return error if any. Besides, return early if OrtRun fails.

The full server log after this fix: 
[server_log.txt](https://github.com/triton-inference-server/onnxruntime_backend/files/7214715/server_log.txt)

The sever will last even if ortrun fails:
```
I0921 20:09:01.361565 1 grpc_server.cc:4111] Started GRPCInferenceService at 0.0.0.0:8001
I0921 20:09:01.361802 1 http_server.cc:2803] Started HTTPService at 0.0.0.0:8000
I0921 20:09:01.403248 1 http_server.cc:162] Started Metrics Service at 0.0.0.0:8002
2021-09-21 20:09:26.315850258 [E:onnxruntime:, sequential_executor.cc:339 Execute] Non-zero status code returned while running Gather node. Name:'bert/embeddings/GatherV2' Status Message: indices element out of data bounds, idx=-1420042007188224409 must be within the inclusive range [-30522,30521]
```

However, if it returns early and then I close the server, the sever will complain as follows:
```
I0921 20:24:49.417818 1 server.cc:249] Timeout 29: Found 1 live models and 0 in-flight non-inference requests
I0921 20:24:50.417946 1 server.cc:249] Timeout 28: Found 1 live models and 0 in-flight non-inference requests
I0921 20:24:51.418066 1 server.cc:249] Timeout 27: Found 1 live models and 0 in-flight non-inference requests
...
I0921 20:25:16.421283 1 server.cc:249] Timeout 2: Found 1 live models and 0 in-flight non-inference requests
I0921 20:25:17.421419 1 server.cc:249] Timeout 1: Found 1 live models and 0 in-flight non-inference requests
I0921 20:25:18.421544 1 server.cc:249] Timeout 0: Found 1 live models and 0 in-flight non-inference requests
E0921 20:25:18.421589 1 main.cc:1571] failed to stop server: Internal - Exit timeout expired. Exiting immediately.
```
Since server does have a live model which failed, is this expected? If not, perhaps there is more action (like closing the live model) for returning early.

Update: before early return, add request release to let server close gracefully

**Motivation**
https://github.com/triton-inference-server/onnxruntime_backend/issues/69 A ORT inference causes Triton server to crash. Although ORT inference fails, the server should not crash and last. I can repro this kind of crash while testing certain bert model with Gather node error (indices element out of data bounds).

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -651,7 +651,7 @@ class ModelInstanceState : public BackendModelInstance {
       const std::string& control_kind, bool required, bool* have_control);
   TRITONSERVER_Error* ValidateInputs(const size_t expected_input_cnt);
   TRITONSERVER_Error* ValidateOutputs();
-  void OrtRun(
+  TRITONSERVER_Error* OrtRun(
       std::vector<TRITONBACKEND_Response*>* responses,
       const uint32_t response_count,
       const std::vector<const char*>& input_names,
@@ -1295,7 +1295,8 @@ ModelInstanceState::ProcessRequests(
   SET_TIMESTAMP(compute_start_ns);
 
   // Run...
-  OrtRun(&responses, request_count, input_names, output_names);
+  RESPOND_ALL_AND_RETURN_IF_ERROR(
+    &responses, request_count, OrtRun(&responses, request_count, input_names, output_names));
 
   uint64_t compute_end_ns = 0;
   SET_TIMESTAMP(compute_end_ns);
@@ -1342,25 +1343,14 @@ ModelInstanceState::ProcessRequests(
       "failed reporting batch request statistics");
 }
 
-void
+TRITONSERVER_Error*
 ModelInstanceState::OrtRun(
     std::vector<TRITONBACKEND_Response*>* responses,
     const uint32_t response_count, const std::vector<const char*>& input_names,
     const std::vector<const char*>& output_names)
 {
-  OrtStatus* status = nullptr;
-  status = ort_api->RunWithBinding(session_, runOptions_, io_binding_);
-  if (status != nullptr) {
-    OrtErrorCode code = ort_api->GetErrorCode(status);
-    std::string msg = ort_api->GetErrorMessage(status);
-    SendErrorForResponses(
-        responses, response_count,
-        TRITONSERVER_ErrorNew(
-            TRITONSERVER_ERROR_INTERNAL,
-            (std::string("onnxruntime execute failure ") +
-             std::to_string(code) + ": " + msg)
-                .c_str()));
-  }
+  RETURN_IF_ORT_ERROR(ort_api->RunWithBinding(session_, runOptions_, io_binding_));
+  return nullptr;
 }
 
 void
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -656,13 +656,13 @@ class ModelInstanceState : public BackendModelInstance {
       const uint32_t response_count,
       const std::vector<const char*>& input_names,
       const std::vector<const char*>& output_names);
-  void SetInputTensors(
+  TRITONSERVER_Error* SetInputTensors(
       size_t total_batch_size, TRITONBACKEND_Request** requests,
       const uint32_t request_count,
       std::vector<TRITONBACKEND_Response*>* responses,
       BackendInputCollector* collector, std::vector<const char*>* input_names,
       bool* cuda_copy);
-  void SetStringInputTensor(
+  TRITONSERVER_Error* SetStringInputTensor(
       TRITONBACKEND_Request** requests, const uint32_t request_count,
       std::vector<TRITONBACKEND_Response*>* responses, const char* input_name,
       std::vector<const char*>* string_ptrs, bool* cuda_copy);
@@ -672,7 +672,7 @@ class ModelInstanceState : public BackendModelInstance {
       std::vector<TRITONBACKEND_Response*>* responses, char* input_buffer,
       std::vector<const char*>* string_ptrs);
   void FillStringData(std::vector<const char*>* string_ptrs, size_t cnt);
-  void ReadOutputTensors(
+  TRITONSERVER_Error* ReadOutputTensors(
       size_t total_batch_size, const std::vector<const char*>& output_names,
       TRITONBACKEND_Request** requests, const uint32_t request_count,
       std::vector<TRITONBACKEND_Response*>* responses);
@@ -858,7 +858,7 @@ ModelInstanceState::ValidateBooleanSequenceControl(
   std::string tensor_datatype;
   RETURN_IF_ERROR(GetBooleanSequenceControlProperties(
       sequence_batching, model_state_->Name(), control_kind, required,
-      &tensor_name, &tensor_datatype, nullptr, nullptr, nullptr, nullptr,
+      &tensor_name, &tensor_datatype, nullptr, nullptr, nullptr, nullptr
       nullptr, nullptr));
   *have_control = !tensor_name.empty();
   if (*have_control) {
@@ -1238,9 +1238,9 @@ ModelInstanceState::ProcessRequests(
       requests, request_count, &responses, model_state_->TritonMemoryManager(),
       model_state_->EnablePinnedInput(), CudaStream(), nullptr, nullptr, 0,
       HostPolicyName().c_str());
-  SetInputTensors(
+  RESPOND_ALL_AND_RETURN_IF_ERROR(requests, request_count, &responses, SetInputTensors(
       total_batch_size, requests, request_count, &responses, &collector,
-      &input_names, &cuda_copy);
+      &input_names, &cuda_copy));
 
   // Request to retrieve all model outputs. 'output_names' and
   // 'output_tensors_' are parallel vectors and so must be kept in
@@ -1271,16 +1271,15 @@ ModelInstanceState::ProcessRequests(
         output_names.emplace_back(io_name);
         output_tensors_.emplace_back(nullptr);
 
-        RESPOND_ALL_AND_RETURN_IF_ORT_ERROR(
-            &responses, request_count,
-            ort_api->BindOutputToDevice(
+        RESPOND_ALL_AND_RETURN_IF_ORT_ERROR(requests, request_count, &responses,
+          ort_api->BindOutputToDevice(
                 io_binding_, io_name, cpu_allocator_info_));
       }
     }
 
     if (err != nullptr) {
-      SendErrorForResponses(&responses, request_count, err);
       output_names.clear();
+      RESPOND_ALL_AND_RETURN_IF_ERROR(requests, request_count, &responses, err);
     }
   }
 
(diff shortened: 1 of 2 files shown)


tanmayv25 (approved):

tanmayv25:
> However, if it returns early and then I close the server, the sever will complain as follows:
> 
> I0921 20:24:49.417818 1 server.cc:249] Timeout 29: Found 1 live models and 0 in-flight non-inference requests
> I0921 20:24:50.417946 1 server.cc:249] Timeout 28: Found 1 live models and 0 in-flight non-inference requests
> I0921 20:24:51.418066 1 server.cc:249] Timeout 27: Found 1 live models and 0 in-flight non-inference requests
> ...
> I0921 20:25:16.421283 1 server.cc:249] Timeout 2: Found 1 live models and 0 in-flight non-inference requests
> I0921 20:25:17.421419 1 server.cc:249] Timeout 1: Found 1 live models and 0 in-flight non-inference requests
> I0921 20:25:18.421544 1 server.cc:249] Timeout 0: Found 1 live models and 0 in-flight non-inference requests
> E0921 20:25:18.421589 1 main.cc:1571] failed to stop server: Internal - Exit timeout expired. Exiting immediately.
> Since server does have a live model which failed, is this expected? If not, perhaps there is more action (like closing the live model) for returning early.

This is not expected. The model should not be be live if the inference failed and the server is expected to close gracefully.

deadeyegoodwin:
> 
> 
> > However, if it returns early and then I close the server, the sever will complain as follows:
> > I0921 20:24:49.417818 1 server.cc:249] Timeout 29: Found 1 live models and 0 in-flight non-inference requests
> > I0921 20:24:50.417946 1 server.cc:249] Timeout 28: Found 1 live models and 0 in-flight non-inference requests
> > I0921 20:24:51.418066 1 server.cc:249] Timeout 27: Found 1 live models and 0 in-flight non-inference requests
> > ...
> > I0921 20:25:16.421283 1 server.cc:249] Timeout 2: Found 1 live models and 0 in-flight non-inference requests
> > I0921 20:25:17.421419 1 server.cc:249] Timeout 1: Found 1 live models and 0 in-flight non-inference requests
> > I0921 20:25:18.421544 1 server.cc:249] Timeout 0: Found 1 live models and 0 in-flight non-inference requests
> > E0921 20:25:18.421589 1 main.cc:1571] failed to stop server: Internal - Exit timeout expired. Exiting immediately.
> > Since server does have a live model which failed, is this expected? If not, perhaps there is more action (like closing the live model) for returning early.
> 
> This is not expected. The model should not be be live if the inference failed and the server is expected to close gracefully.

"the model should not be live if the inference failed"... Just to be clear. The backend should be able to gracefully unload all models on exit, even if at some point one or more inference requests failed. So need to determine why this is not the case...

tanmayv25:
May its because we are missing releasing the request objects before returning. See how TensorFlow backend does that [here ](https://github.com/triton-inference-server/tensorflow_backend/blob/main/src/tensorflow.cc#L1986) before returning.

jcwchen:
Thank you both for the info! I am taking deeper look at it

jcwchen:
> May its because we are missing releasing the request objects before returning. See how TensorFlow backend does that [here ](https://github.com/triton-inference-server/tensorflow_backend/blob/main/src/tensorflow.cc#L1986) before returning.

Yes, I believe `TRITONBACKEND_RequestRelease` is the key. After adding this after cacthing OrtRun error, now the server can be gracefully closed without any warning/error. Please review this PR again. Thank you for your review!

jcwchen:
One follow-up question: do we need to `TRITONBACKEND_RequestRelease` after every `SendErrorForResponses`? I saw there are other `SendErrorForResponses` but they did not release requests after that.

tanmayv25:
I believe that requests must always be released before returning. Otherwise, we risk ungraceful exits as observed by you. You can add the fix for the cases you have found.
FYI: I used [FAIL_ALL_AND_RETURN_IF_ERROR](https://github.com/triton-inference-server/tensorrt_backend/blob/main/src/tensorrt.cc#L55) macro in tensorrt_backend to ensure the requests get released when returning and also the failed stats are required.

jcwchen:
@tanmayv25 Thank you for the guidance. I have updated `RESPOND_ALL_AND_RETURN_IF_ORT_ERROR`  and `RESPOND_ALL_AND_RETURN_IF_ERROR` to make them release requests after `SendErrorForResponses`.  To accomplish that, I also kind of refactored other functions to catch server error internally and externally return properly. Please review it again. Thanks!

src/onnxruntime.cc
```diff
@@ -858,7 +858,7 @@ ModelInstanceState::ValidateBooleanSequenceControl(
   std::string tensor_datatype;
   RETURN_IF_ERROR(GetBooleanSequenceControlProperties(
       sequence_batching, model_state_->Name(), control_kind, required,
-      &tensor_name, &tensor_datatype, nullptr, nullptr, nullptr, nullptr,
+      &tensor_name, &tensor_datatype, nullptr, nullptr, nullptr, nullptr
```
tanmayv25, line 861: Does this code still compiles? The comma at the end is still needed.

tanmayv25 (changes requested):
Overall looks good but please see my comments. 
Once you confirm the code compiles are runs properly for the failing case, I will initiate the CI.

Just as a FYI: We plan to add RESPOND_ALL_AND_RETURN_IF_ERROR and similar macros in our backend repo so all backends don't have to reimplement this function.
We will initiate an audit of all backends at some point in the future to weed out incorrect returns from ProcessRequest function.

src/onnxruntime.cc
jcwchen: Good catch! Just corrected my serious typo. I only changed this line to make it compatible with my local triton-inference-server version... This updated code now can be compiled

jcwchen:
We plan to add RESPOND_ALL_AND_RETURN_IF_ERROR and similar macros in our backend repo so all backends don't have to reimplement this function.
> We will initiate an audit of all backends at some point in the future to weed out incorrect returns from ProcessRequest function.

Good to know. When referencing other backends, I had the same thought that some macros should go into common utilities under backend/ (just like RETURN_IF_ERROR). I am happy to see the future backend update!

src/onnxruntime_utils.h
```diff
@@ -37,13 +37,50 @@ namespace triton { namespace backend { namespace onnxruntime {
...
-      SendErrorForResponses(RESPONSES, RESPONSES_COUNT, raarie_err__); \
-      return;                                                          \
-    }                                                                  \
+#define RESPOND_ALL_AND_RETURN_IF_ERROR(                            \
+    REQUESTS, REQUEST_COUNT, RESPONSES, S)                          \
+  do {                                                              \
+    TRITONSERVER_Error* raarie_err__ = (S);                         \
+    if (raarie_err__ != nullptr) {                                  \
+      SendErrorForResponses(RESPONSES, REQUEST_COUNT, raarie_err__);\
+      for (uint32_t r = 0; r < REQUEST_COUNT; ++r) {                \
```
askhade, line 46: nit: you should create a utility function similar to SendErrorForResponses since this for loop is present in 2 places in this file

src/onnxruntime_utils.h
```diff
@@ -37,13 +37,50 @@ namespace triton { namespace backend { namespace onnxruntime {
...
+    if (status__ != nullptr) {                                             \
+      OrtErrorCode code = ort_api->GetErrorCode(status__);                 \
+      std::string msg = std::string(ort_api->GetErrorMessage(status__));   \
+      ort_api->ReleaseStatus(status__);                                    \
+      auto err__ = TRITONSERVER_ErrorNew(                                  \
+              TRITONSERVER_ERROR_INTERNAL,                                 \
+              (std::string("onnx runtime error ") + std::to_string(code) + \
+               ": " + msg)                                                 \
+                  .c_str());                                               \
+      SendErrorForResponses(RESPONSES, REQUEST_COUNT, err__);              \
```
askhade, line 72: nit: 
```
SendErrorForResponses(
          RESPONSES, REQUEST_COUNT, 
          TRITONSERVER_ErrorNew(
              TRITONSERVER_ERROR_INTERNAL,
              (std::string("onnx runtime error ") + std::to_string(code) +
               ": " + msg)
                  .c_str()));    
```

src/onnxruntime_utils.h
```diff
@@ -37,13 +37,50 @@ namespace triton { namespace backend { namespace onnxruntime {
...
-      return;                                                          \
-    }                                                                  \
+#define RESPOND_ALL_AND_RETURN_IF_ERROR(                            \
+    REQUESTS, REQUEST_COUNT, RESPONSES, S)                          \
+  do {                                                              \
+    TRITONSERVER_Error* raarie_err__ = (S);                         \
+    if (raarie_err__ != nullptr) {                                  \
+      SendErrorForResponses(RESPONSES, REQUEST_COUNT, raarie_err__);\
+      for (uint32_t r = 0; r < REQUEST_COUNT; ++r) {                \
+        if (RESPONSES[r] != nullptr) {                              \
```
askhade, line 47: not sure how this works... 
SendErrorForResponses sets walks through each response and after response is sent sets it to nullptr

Check: https://github.com/triton-inference-server/tensorrt_backend/blob/main/src/tensorrt.cc#L93

tanmayv25: I believe it worked because dynamic batching was not being used. @jcwchen  Can you address @askhade's comment and verify your changes work even when dynamic batching is enabled?

jcwchen: Trying. How can I enable dynamic batching? Follow the steps here? https://github.com/triton-inference-server/server/blob/main/docs/architecture.md Thanks

tanmayv25: So, yes.. Adding  `dynamic_batching { max_queue_delay_microseconds: 5000000}` in the model config will allow Triton to wait for 5sec for forming batches. 

Just make sure that you would need to send multiple inference requests to the server within the time interval so that they can be batched together. Best way would be to use perf_analyzer with --concurrency-range=<max_batch_size of the model>.

I would recommend keeping the max_batch_size of the model greater than or equal to 4.

jcwchen: Thank you for the detailed instructions! Although I have tried other models (which did not crash before) with dynamic batching and every thing looks good, I cannot enable dynamic batching with my crashed model (it seems it does not support max_batch_size and always complains the shape is wrong)...  It's quite hard for me to verify my update with dynamic batching, but it can still resolve the crash issue without enabling dynamic batching for sure.

Thanks @askhade for pointing out and I think you are right so I have made send response error + request release be same as the reference code; also refactored the structure to shared the common code. Please review it again

askhade (approved):

askhade:
@tanmayv25 : PR looks good to me. Please approve if it looks good to you too.

tanmayv25 (approved):
Looks good to me as well. I have started a CI run with the latest changes.

jcwchen:
@tanmayv25 How is the CI going? Could we forward this PR before the code freeze? Thanks

tanmayv25:
The build failed for some reason. I am trying to root cause it and will let you know if I find something off.

tanmayv25:
The CI passed.
FYI: there was some issue with unavailability of eigen repository which caused the build to fail.

PR #72, opened 2021-09-27
enable ort configs

askhade:
1. Update ORT version to 1.9
2. Enable configuring cuda and cpu provider specific optimization options for ORT backend.
3. Update Tensorrt calibration table default value to support Q\DQ models.


src/onnxruntime.cc
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright (c) 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights
```
deadeyegoodwin, line 1: Remove the '(c)' so that the formatter doesn't wrap this line

src/onnxruntime.cc
```diff
@@ -339,17 +358,42 @@ ModelState::LoadModel(
...
-      };
+      cuda_options.default_memory_arena_cfg = nullptr;
+
+      {
+        // Parse CUDA EP configurations
+        triton::common::TritonJson::Value params;
+        if (model_config_.Find("parameters", &params)) {
+          int cudnn_conv_algo_search = 0;
+          RETURN_IF_ERROR(TryParseParameter(
+              params, "cudnn_conv_algo_search", &cudnn_conv_algo_search));
```
deadeyegoodwin, line 380: You are assuming 'cudnn_conv_algo_search' is unmodified by this call (and similar below). That is not documented and seems like an unsafe API design. GetParameterValue and underlying calls do not make any guarantees about what happens to that argument when there is an error.

src/onnxruntime_utils.h
```diff
@@ -158,4 +159,34 @@ TRITONSERVER_Error* CompareDimsSupported(
     const std::vector<int64_t>& model_shape, const std::vector<int64_t>& dims,
     const int max_batch_size, const bool compare_exact);
 
+/// Parse the string as unsigned long integer.
+///
+/// \param value The string.
+/// \param parse_value The unsigned long integral value of the string.
+/// \return a TRITONSERVER_Error indicating success or failure.
+TRITONSERVER_Error* ParseUnsignedLongValue(
```
deadeyegoodwin, line 167: Can this be ParseUnsignedIntValue and you can add it to backend_common.cc along with the other parse utility functions.

src/onnxruntime_utils.h
```diff
@@ -158,4 +159,34 @@ TRITONSERVER_Error* CompareDimsSupported(
...
+    const std::string& value, unsigned long* parsed_value);
+
+/// Try to parse the requested parameter.
+///
+/// \param params The param in model config
+/// \param value The parsed string value.
+/// \param mkey Key in the model config.
+/// \return a TRITONSERVER_Error indicating success or failure. Ignores
+/// TRITONSERVER_ERROR_NOT_FOUND
+TRITONSERVER_Error* TryParseParameter(
```
deadeyegoodwin, line 177: Can you add these to backend_common.cc and change them to have a default argument that is used to set 'value' when the key is not found. That will remove any ambiquity about what happens to 'value' when the key is not found.

README.md
```diff
@@ -77,7 +77,7 @@ TensorRT and OpenVino support:
 ```
 $ mkdir build
 $ cd build
-$ cmake -DCMAKE_INSTALL_PREFIX:PATH=`pwd`/install -DTRITON_BUILD_ONNXRUNTIME_VERSION=1.8.1 -DTRITON_BUILD_CONTAINER_VERSION=21.07 -DTRITON_ENABLE_ONNXRUNTIME_TENSORRT=ON -DTRITON_ENABLE_ONNXRUNTIME_OPENVINO=ON -DTRITON_BUILD_ONNXRUNTIME_OPENVINO_VERSION=2021.2.200 ..
+$ cmake -DCMAKE_INSTALL_PREFIX:PATH=`pwd`/install -DTRITON_BUILD_ONNXRUNTIME_VERSION=1.9.0 -DTRITON_BUILD_CONTAINER_VERSION=21.08 -DTRITON_ENABLE_ONNXRUNTIME_TENSORRT=ON -DTRITON_ENABLE_ONNXRUNTIME_OPENVINO=ON -DTRITON_BUILD_ONNXRUNTIME_OPENVINO_VERSION=2021.2.200 ..
```
deadeyegoodwin, line 80: Are you changing ORT version to 1.9.0 in the 21.10 release? If so can you do that in a separate PR and be sure to update build.py in the server repo as well.

deadeyegoodwin (changes requested):

src/onnxruntime.cc
askhade: I updated the helper functions to use the default value when key is not found. In case of any other error (other than key not found) the code returns. I intentionally do not do any error checking for these values example are the values within allowed range etc because ORT does that... In case the value is not set right then ORT session creation will fail.

README.md
askhade: As discussed in today's meeting I will use this same PR for ort version update as well. I updated the PR description to reflect the same. 

https://github.com/triton-inference-server/server/pull/3412 server PR for version update plus tests

src/onnxruntime.cc
```diff
@@ -1,4 +1,5 @@
-// Copyright (c) 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+// Copyright 2019-2021, NVIDIA CORPORATION & AFFILIATES. All rights
```
deadeyegoodwin, line 1: Fix to be single line... otherwise our copyright checker will fail

deadeyegoodwin (changes requested):

deadeyegoodwin (commented):
Looks good. Will run CI

PR #79, opened 2021-10-29
update tensorrt parser to support TRT 8.2

askhade:
Triton 21.12 will include TRT 8.2. This PR is in preparation for this change.

With this change we let the user to provide a branch\tag for onnx-tensorrt parser repo "TRITON_ONNX_TENSORRT_REPO_TAG "

If this tag is specified we simply use this to build ORT. If nothing is specified then we use the defaults listed in ORT_TO_TRTPARSER_VERSION_MAP specific to the ort and trt version being built. If this map does not have an entry for the ort version and trt version combo then we simply use the default branch\tag which ort uses.

This is to enable support for versions of TRT which ORT release does not support yet. 

Note: I have tested this on linux but not windows.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,13 +58,19 @@ project(tritononnxruntimebackend LANGUAGES C CXX)
 #         TRITON_BUILD_CUDA_HOME="C:\Program Files\NVIDIA GPU Computing Toolkit\v11.1"
 #
 #   - If you want TensorRT support set
-#     TRITON_ENABLE_ONNXRUNTIME_TENSORRT=ON and set
-#     TRITON_BUILD_TENSORRT_HOME.
+#     TRITON_ENABLE_ONNXRUNTIME_TENSORRT=ON and set TRITON_BUILD_TENSORRT_HOME.
+#
 #     Optionally set TRITON_ONNX_TENSORRT_REPO_TAG to specify a branch in https://github.com/onnx/onnx-tensorrt repo
 #     example:
 #         TRITON_ONNX_TENSORRT_REPO_TAG=master
 #     This enables using a version of tensorrt which is not yet supported in ONNXRuntime release branch.
 #     By default we pick the default branch which comes with the requested version of onnxruntime.
+#
+#     Optionally set TRITON_TENSORRT_VERSION to specify the verison of TRT which is being used. 
+#     This along with TRITON_BUILD_ONNXRUNTIME_VERSION is used to pick the right onnx tensorrt parser version. 
+#     When TRITON_ONNX_TENSORRT_REPO_TAG is set TRITON_TENSORRT_VERSION is ignored. 
+#     When neither TRITON_ONNX_TENSORRT_REPO_TAG or TRITON_TENSORRT_VERSION are set 
+#     the default parser version which comes with ORT is picked.
 #
 #   - If you want OpenVINO support set
 #     TRITON_ENABLE_ONNXRUNTIME_OPENVINO=ON and set
@@ -91,6 +97,7 @@ set(TRITON_BUILD_CUDNN_HOME "" CACHE PATH "Path to CUDNN install")
 set(TRITON_BUILD_TENSORRT_HOME "" CACHE PATH "Path to TensorRT install")
 set(TRITON_ONNXRUNTIME_INCLUDE_PATHS "" CACHE PATH "Paths to ONNXRuntime includes")
 set(TRITON_ONNX_TENSORRT_REPO_TAG "" CACHE STRING "Tag for onnx-tensorrt repo")
+set(TRITON_TENSORRT_VERSION "" CACHE STRING "TRT version for this build.")
 set(TRITON_ONNXRUNTIME_LIB_PATHS "" CACHE PATH "Paths to ONNXRuntime libraries")
 
 set(TRITON_BACKEND_REPO_TAG "main" CACHE STRING "Tag for triton-inference-server/backend repo")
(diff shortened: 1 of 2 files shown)


tools/gen_ort_dockerfile.py
```diff
@@ -32,6 +32,13 @@
 
 FLAGS = None
 
+ORT_TO_TRTPARSER_VERSION_MAP = {
+    '1.9.1': (
+        '21.12',         # Triton container version
+        'release/8.2-EA' # ONNX-Tensorrt parser version
```
jywu-msft, line 38: I think we can wait for the GA version of the parser to merge this?
EA was primarily for us to start testing earlier. 
@kevinch-nv , I assume there will be an 8.2 ga branch created in the next week or two?

askhade: Right! this PR is meant for us to start testing for 21.12 - Triton-ORT-TRT8.2

kevinch-nv: Yes, the GA branch will be created once the 8.2 GA is pubicaly released.

deadeyegoodwin:
I converted this PR to a draft so it doesn't accidentally get merged

CMakeLists.txt
```diff
@@ -60,6 +60,11 @@ project(tritononnxruntimebackend LANGUAGES C CXX)
 #   - If you want TensorRT support set
 #     TRITON_ENABLE_ONNXRUNTIME_TENSORRT=ON and set
 #     TRITON_BUILD_TENSORRT_HOME.
+#     Optionally set TRITON_ONNX_TENSORRT_REPO_TAG to branch in https://github.com/onnx/onnx-tensorrt repo
```
deadeyegoodwin, line 63: "... to specify a branch in..."

tools/gen_ort_dockerfile.py
```diff
@@ -498,6 +520,16 @@ def preprocess_gpu_flags():
     if FLAGS.enable_gpu:
         preprocess_gpu_flags()
 
+    # if a tag is provided by the user, then simply use it
+    # if the tag is empty - check whether there is an entry in the ORT_TO_TRTPARSER_VERSION_MAP
+    # map cprresponding to ort version + triton container version combo. If yes then use it
```
deadeyegoodwin, line 525: corresponding

deadeyegoodwin (changes requested):

tools/gen_ort_dockerfile.py
```diff
@@ -32,6 +32,13 @@
 
 FLAGS = None
 
+ORT_TO_TRTPARSER_VERSION_MAP = {
+    '1.9.1': (
+        '21.12',         # Triton container version
```
deadeyegoodwin, line 37: Do we need the Triton container version? Can't we just have a map from ORT version to corresponding parser version?

deadeyegoodwin (changes requested):

tools/gen_ort_dockerfile.py
askhade: No it is not enough... we either need triton version or trt version... because which version of parser to pick depends on the trt version. I chose to use triton container version instead of trt version here because we already have the info regarding triton version so simply using it is easy.

deadeyegoodwin: The problem is that linux x86, linux arm, jetson and windows may use different versions of TensorRT in each release. I think we need to extract the TRT version from the image in build.py and then pass it to ORT backend build by another ARG. Then your logic would be as it is now except using TRT version (if given) instead of Triton version. I can make the build.py changes to determine the TRT version.

deadeyegoodwin (changes requested):

tools/gen_ort_dockerfile.py
askhade: ok makes sense. I will update this PR to use trt version instead...

tools/gen_ort_dockerfile.py
```diff
@@ -206,23 +221,23 @@ def dockerfile_for_linux(output_file):
        /opt/onnxruntime/include
 
 RUN mkdir -p /opt/onnxruntime/lib && \
-    cp /workspace/build/Release/libonnxruntime_providers_shared.so \
+    cp /workspace/build/${ONNXRUNTIME_BUILD_CONFIG}/libonnxruntime_providers_shared.so \
```
deadeyegoodwin, line 224: Does this PR need to be rebased? It seems to have the debug build changes. Can we merge the debug changes and rebase this? Or are debug changes not ready yet?

deadeyegoodwin:
I'll working on updating build.py to provide TRITON_TENSORRT_VERSION to the onnx runtime build.

tools/gen_ort_dockerfile.py
askhade: Rebase Done.

deadeyegoodwin (approved):

tools/gen_ort_dockerfile.py
```diff
@@ -32,6 +32,13 @@
 
 FLAGS = None
 
+ORT_TO_TRTPARSER_VERSION_MAP = {
+    '1.9.1': (
+        '8.2',         # TensorRT version
```
deadeyegoodwin, line 37: The TensorRT versions that you will see will be a.b.c.d or even something like 8.0.3.4+cuda11.3.1.005
So I think you will need to parse --trt-version to look only at a.b portion of the version

deadeyegoodwin (changes requested):

deadeyegoodwin:
@askhade 
server PR: https://github.com/triton-inference-server/server/pull/3588

Here is how onnxruntime_backend cmake will be invoked once PR is merged

cmake ['-DTRITON_BUILD_ONNXRUNTIME_VERSION=1.9.0', '-DTRITON_ENABLE_ONNXRUNTIME_TENSORRT=ON', '-DTRITON_BUILD_CONTAINER=gitlab-master.nvidia.com:5005/dl/dgx/cuda:11.4-devel-ubuntu20.04--21.10', '-DTRITON_ENABLE_ONNXRUNTIME_OPENVINO=ON', '-DTRITON_BUILD_ONNXRUNTIME_OPENVINO_VERSION=2021.2.200', '-DCMAKE_BUILD_TYPE=Release', '-DCMAKE_INSTALL_PREFIX:PATH=/tmp/tritonbuild/onnxruntime/install', '-DTRITON_COMMON_REPO_TAG:STRING=main', '-DTRITON_CORE_REPO_TAG:STRING=main', '-DTRITON_BACKEND_REPO_TAG:STRING=main', '-DTRITON_ENABLE_GPU:BOOL=ON', '-DTRITON_ENABLE_MALI_GPU:BOOL=OFF', '-DTRT_VERSION=8.0.3.4+cuda11.3.1.005', '-DDALI_VERSION=1.6.0', '..']

askhade:
> @askhade server PR: [triton-inference-server/server#3588](https://github.com/triton-inference-server/server/pull/3588)
> 
> Here is how onnxruntime_backend cmake will be invoked once PR is merged
> 
> cmake ['-DTRITON_BUILD_ONNXRUNTIME_VERSION=1.9.0', '-DTRITON_ENABLE_ONNXRUNTIME_TENSORRT=ON', '-DTRITON_BUILD_CONTAINER=gitlab-master.nvidia.com:5005/dl/dgx/cuda:11.4-devel-ubuntu20.04--21.10', '-DTRITON_ENABLE_ONNXRUNTIME_OPENVINO=ON', '-DTRITON_BUILD_ONNXRUNTIME_OPENVINO_VERSION=2021.2.200', '-DCMAKE_BUILD_TYPE=Release', '-DCMAKE_INSTALL_PREFIX:PATH=/tmp/tritonbuild/onnxruntime/install', '-DTRITON_COMMON_REPO_TAG:STRING=main', '-DTRITON_CORE_REPO_TAG:STRING=main', '-DTRITON_BACKEND_REPO_TAG:STRING=main', '-DTRITON_ENABLE_GPU:BOOL=ON', '-DTRITON_ENABLE_MALI_GPU:BOOL=OFF', '-DTRT_VERSION=8.0.3.4+cuda11.3.1.005', '-DDALI_VERSION=1.6.0', '..']

Thanks @deadeyegoodwin . I will fix the PR to correctly pick up the version.

GuanLuo (approved):

deadeyegoodwin (approved):

PR #83, opened 2021-11-20
remove version dependency on libonnxruntime.so

askhade:
Right now libtriton_onnxruntime.so has a strict version dependency on libonnxruntime.so. This makes it difficult to replace ort binaries in released containers for easy experimentation. Current solution is to build triton-ort backend which is not always possible.

### DIFF
--- a/tools/gen_ort_dockerfile.py
+++ b/tools/gen_ort_dockerfile.py
@@ -219,10 +219,8 @@ RUN mkdir -p /opt/onnxruntime/include && \
 RUN mkdir -p /opt/onnxruntime/lib && \
     cp /workspace/build/${ONNXRUNTIME_BUILD_CONFIG}/libonnxruntime_providers_shared.so \
        /opt/onnxruntime/lib && \
-    cp /workspace/build/${ONNXRUNTIME_BUILD_CONFIG}/libonnxruntime.so.${ONNXRUNTIME_VERSION} \
-       /opt/onnxruntime/lib && \
-    (cd /opt/onnxruntime/lib && \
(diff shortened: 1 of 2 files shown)


tools/gen_ort_dockerfile.py
```diff
@@ -183,6 +183,14 @@ def dockerfile_for_linux(output_file):
     df += '''
 WORKDIR /workspace/onnxruntime
 ARG COMMON_BUILD_ARGS="--config ${ONNXRUNTIME_BUILD_CONFIG} --skip_submodule_sync --parallel --build_shared_lib --build_dir /workspace/build --cmake_extra_defines CMAKE_CUDA_ARCHITECTURES='52;60;61;70;75;80;86' "
+'''
+
+    # Remove version info from libonnxruntime.so
+    # This makes it possible to replace ort binaries in released triton containers
+    # for experimentation, without having to build triton-ort backend.
+    df += '''
+RUN sed -i 's/VERS_%s//' tools/ci_build/gen_def.py &&  (sed -i 's/% VERSION_STRING//' tools/ci_build/gen_def.py)
```
deadeyegoodwin, line 192: Do you need to do something for Windows?

askhade: No this is specific to linux

askhade: Right now we are running into this issue for linux deployments and A/B testing. This specific solution is specific to linux... I havent tested if this is a problem for windows... This PR targets windows only

deadeyegoodwin (approved):

deadeyegoodwin:
I'll run CI for this change and then commit.

PR #98, opened 2022-02-08
enable io binding for outputs

askhade:
This PR enables io-binding for outputs.

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -1488,7 +1488,7 @@ ModelInstanceState::ProcessRequests(
 
         if (err != nullptr) {
           LOG_MESSAGE(
-              TRITONSERVER_LOG_WARN,
+              TRITONSERVER_LOG_VERBOSE,
               (std::string("Output Properties Unavailable. Using cpu as "
                            "preferred location for all outputs.")
                    .c_str()));
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -1490,18 +1490,22 @@ ModelInstanceState::ProcessRequests(
 
         // save this as we need the mem type and device id for reading the
         // outputs.
-        output_device_info_[output_name] = {memory_type, memory_type_id};
+        output_device_info_.insert(
+            {output_name.first, {memory_type, memory_type_id}});
       } else {
-        auto output_device_info_iter = output_device_info_.find(name);
+        auto output_device_info_iter =
+            output_device_info_.find(output_name.first);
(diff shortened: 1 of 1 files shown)


deadeyegoodwin:
@tanmayv25  and @GuanLuo, please review for 22.02.
@askhade I will start a CI with this change.

tanmayv25:
Looks good to me. @askhade  just update the copyright year at the top.

GuanLuo (dismissed):

src/onnxruntime.cc
```diff
@@ -1453,17 +1458,69 @@ ModelInstanceState::ProcessRequests(
...
+          RESPOND_ALL_AND_SET_TRUE_IF_ERROR(
+              responses, request_count, all_response_failed,
+              TRITONSERVER_ErrorNew(
+                  TRITONSERVER_ERROR_INTERNAL,
+                  (std::string("device info for output tensor '") +
+                   output_name.first + "' not found")
+                      .c_str()));
+        }
+        memory_type = output_device_info_iter->second.first;
+        memory_type_id = output_device_info_iter->second.second;
```
GuanLuo, line 1515: I am fine with using CPU for default, or you may use the `preferred_memory_type/id` that you defined earlier, if "output is preferred to be close to where the model is deployed" is a valid assumption

askhade: I considered this option. The reason I did not go with it because there could be perf implications. The same reason why we chose to keep outputs on cpu by default before this query api was implemented. 
For http/grpc non ensemble scenarios if the output is on cpu then multiple copies will be done ORT -> CPU to GPU and then Triton GPU -> CPU .

src/onnxruntime.cc
```diff
@@ -1453,17 +1458,69 @@ ModelInstanceState::ProcessRequests(
...
+        // to this destination memory type. The destination memory type
+        // for an output for all requests should be same. So use any request for
+        // this query.
+        auto err = TRITONBACKEND_RequestOutputBufferProperties(
+            requests[0], output_name.first.c_str(), /*byte_size*/ nullptr,
+            &memory_type, &memory_type_id);
+
+        if (err != nullptr) {
+          LOG_MESSAGE(
+              TRITONSERVER_LOG_WARN,
```
deadeyegoodwin, line 1491: Make this verbose, otherwise you will get a warning for every request to the model.

GuanLuo (approved):

GuanLuo (approved):

PR #108, opened 2022-03-21
Fix L0_infer_cudashm failure

Tabrizian:
The root cause for the segfault was that when instance kind is KIND_CPU and the OutputBufferQuery returns a GPU memory, `cuda_allocator_info_` is a nullptr and will lead to a segfault. Stacktrace of the segfault:

```
#1  0x00007f8338fa44d9 in triton::backend::onnxruntime::ModelInstanceState::ProcessRequests (this=0x7f80d80115c0, requests=0x7f81580440c0, request_count=1) at /workspace/onnxruntime_backend/src/onnxruntime.cc:1630
#2  0x00007f8338fb10c6 in triton::backend::onnxruntime::TRITONBACKEND_ModelInstanceExecute (instance=0x7f80d8015c30, requests=0x7f81580440c0, request_count=1) at /workspace/onnxruntime_backend/src/onnxruntime.cc:2600
```

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -1547,9 +1547,6 @@ ModelInstanceState::ProcessRequests(
     if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
       preferred_memory_type = TRITONSERVER_MEMORY_GPU;
       preferred_memory_type_id = DeviceId();
-    } else {
-      preferred_memory_type = TRITONSERVER_MEMORY_CPU;
-      preferred_memory_type_id = 0;
     }
 
     // Request to retrieve all model outputs. 'output_names' and
@@ -1601,8 +1598,8 @@ ModelInstanceState::ProcessRequests(
           }
         }
 
-        // If the instance type is not KIND_GPU, set the memory type to CPU.
-        if (Kind() != TRITONSERVER_INSTANCEGROUPKIND_GPU) {
+        // If the cuda allocator is not set, bind the output to CPU.
+        if (cuda_allocator_info_ == nullptr) {
           memory_type = TRITONSERVER_MEMORY_CPU;
           memory_type_id = 0;
         }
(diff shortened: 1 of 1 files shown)


src/onnxruntime.cc
```diff
@@ -1547,6 +1547,9 @@ ModelInstanceState::ProcessRequests(
     if (Kind() == TRITONSERVER_INSTANCEGROUPKIND_GPU) {
       preferred_memory_type = TRITONSERVER_MEMORY_GPU;
       preferred_memory_type_id = DeviceId();
+    } else {
+      preferred_memory_type = TRITONSERVER_MEMORY_CPU;
+      preferred_memory_type_id = 0;
```
GuanLuo, line 1552: You don't need to add this, in line 1545 the values have been initialized to CPU / 0

src/onnxruntime.cc
```diff
@@ -1598,6 +1600,13 @@ ModelInstanceState::ProcessRequests(
             memory_type_id = 0;
           }
         }
+
+        // If the instance type is not KIND_GPU, set the memory type to CPU.
+        if (Kind() != TRITONSERVER_INSTANCEGROUPKIND_GPU) {
```
GuanLuo, line 1605: Can you check if `cuda_allocator_info_` is initialized instead? (cuda_allocator_info_ == nullptr). That is the direct cause of can't put output to device memory, and I would imagine the use case where the model is deployed on CPU but have `cuda_allocator_info_` set so that ORT will place the output to device memory if desired.

GuanLuo (changes requested):

src/onnxruntime.cc
Tabrizian: Sounds good. Fixed.

GuanLuo (approved):

Tabrizian:
The L0_batcher test failed in the CI. I verified it on a DLCluster machine and it passed. Merging this PR.

PR #112, opened 2022-04-19
Default to load model on GPU when instance kind is not specified 

CoderHam:
and autocomplete is used

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -607,6 +607,48 @@ ModelState::LoadModel(
         }
       }
     }
+  } else {
+    // If instance kind is AUTO, default to use GPU execution provider.
+    // Using default values for everything other than device id and cuda
+    // stream
+    OrtCUDAProviderOptions cuda_options;
+    cuda_options.device_id = instance_group_device_id;
+    cuda_options.has_user_compute_stream = stream != nullptr ? 1 : 0;
+    cuda_options.user_compute_stream =
+        stream != nullptr ? (void*)stream : nullptr,
+    cuda_options.default_memory_arena_cfg = nullptr;
+
+    {
+      // Parse CUDA EP configurations
+      triton::common::TritonJson::Value params;
+      if (model_config_.Find("parameters", &params)) {
+        int cudnn_conv_algo_search = 0;
+        RETURN_IF_ERROR(TryParseModelStringParameter(
+            params, "cudnn_conv_algo_search", &cudnn_conv_algo_search, 0));
+        cuda_options.cudnn_conv_algo_search =
+            static_cast<OrtCudnnConvAlgoSearch>(cudnn_conv_algo_search);
+
+        RETURN_IF_ERROR(TryParseModelStringParameter(
+            params, "gpu_mem_limit", &cuda_options.gpu_mem_limit,
+            std::numeric_limits<size_t>::max()));
+
+        RETURN_IF_ERROR(TryParseModelStringParameter(
+            params, "arena_extend_strategy",
+            &cuda_options.arena_extend_strategy, 0));
+
+        RETURN_IF_ERROR(TryParseModelStringParameter(
+            params, "do_copy_in_default_stream",
+            &cuda_options.do_copy_in_default_stream, true));
+      }
+    }
+
+    RETURN_IF_ORT_ERROR(ort_api->SessionOptionsAppendExecutionProvider_CUDA(
+        soptions, &cuda_options));
+    LOG_MESSAGE(
+        TRITONSERVER_LOG_VERBOSE,
+        (std::string("CUDA Execution Accelerator is set for '") + Name() +
+         "' on device " + std::to_string(instance_group_device_id))
+            .c_str());
   }
 
   // Register all op libraries that contain custom operations.


src/onnxruntime.cc
```diff
@@ -608,6 +608,51 @@ ModelState::LoadModel(
       }
     }
   }
+#ifdef TRITON_ENABLE_GPU
+  else {
```
tanmayv25: What about cases where the config provides the execution accelerators?  We are skipping adding those.

> I need to do some additional cleanup but as I understand the check at [https://github.com/triton-inference-server/onnxruntime_backend/blob/90c09650d190fd83f58fd4b1039acdc3092180fb/src/onnxruntime.cc#L391](https://nam11.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Ftriton-inference-server%2Fonnxruntime_backend%2Fblob%2F90c09650d190fd83f58fd4b1039acdc3092180fb%2Fsrc%2Fonnxruntime.cc%23L391&data=05%7C01%7Ctanmayv%40nvidia.com%7C94fbd1bddeeb4660d77808da223a6076%7C43083d15727340c1b7db39efd9ccc17a%7C0%7C0%7C637859931363782285%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=PEDfEkGdjFMDX6HJJEv62zE0%2FOVosC5PkMMGBfB4V0A%3D&reserved=0) is not needed and we can change L397 from (instance_group_kind == TRITONSERVER_INSTANCEGROUPKIND_GPU) to (instance_group_kind != TRITONSERVER_INSTANCEGROUPKIND_CPU)

Why are we not making the change you said we should here?

CoderHam: Cleaned up. Please look at latest changes

src/onnxruntime.cc
```diff
@@ -388,221 +388,219 @@ ModelState::LoadModel(
   // to indicate that execution providers should not be added (this is
```
tanmayv25, line 388: Would need to update this comment to describe new behavior where we add CUDA EP for kind == AUTO.

src/onnxruntime.cc
```diff
@@ -388,221 +388,219 @@ ModelState::LoadModel(
...
-        if (optimization.Find("execution_accelerators", &eas)) {
-          triton::common::TritonJson::Value cpu_eas;
-          if (eas.Find("cpu_execution_accelerator", &cpu_eas)) {
-            for (size_t ea_idx = 0; ea_idx < cpu_eas.ArraySize(); ea_idx++) {
-              triton::common::TritonJson::Value ea;
-              RETURN_IF_ERROR(cpu_eas.IndexAsObject(ea_idx, &ea));
-              std::string name;
-              RETURN_IF_ERROR(ea.MemberAsString("name", &name));
+  // CPU execution providers
+  if (instance_group_kind != TRITONSERVER_INSTANCEGROUPKIND_AUTO) {
```
tanmayv25, line 568: I don't think this condition is required. If the config provides `cpu_execution_accelerator` then only this will get added.
For KIND_AUTO, the config might still provide a CPU accelerator and this condition would render it ineffective.

src/onnxruntime.cc
```diff
@@ -388,221 +388,219 @@ ModelState::LoadModel(
...
-        triton::common::TritonJson::Value eas;
-        if (optimization.Find("execution_accelerators", &eas)) {
-          triton::common::TritonJson::Value gpu_eas;
-          if (eas.Find("gpu_execution_accelerator", &gpu_eas)) {
-            for (size_t ea_idx = 0; ea_idx < gpu_eas.ArraySize(); ea_idx++) {
-              triton::common::TritonJson::Value ea;
-              RETURN_IF_ERROR(gpu_eas.IndexAsObject(ea_idx, &ea));
-              std::string name;
-              RETURN_IF_ERROR(ea.MemberAsString("name", &name));
+  if ((instance_group_kind == TRITONSERVER_INSTANCEGROUPKIND_GPU) &&
```
tanmayv25, line 396: shouldn't this be:
`(instance_group_kind == TRITONSERVER_INSTANCEGROUPKIND_GPU) || (instance_group_kind == TRITONSERVER_INSTANCEGROUPKIND_AUTO)`
?
Even with kind_auto, the config may specify TensorRT execution provider optimization.

src/onnxruntime.cc
```diff
@@ -388,221 +388,219 @@ ModelState::LoadModel(
...
+      (instance_group_kind != TRITONSERVER_INSTANCEGROUPKIND_AUTO)) {
```
tanmayv25, line 397: Keep CUDA_EP in the same condition block.

tanmayv25 (changes requested):

src/onnxruntime.cc
CoderHam: We do not want to add execution accelerators when it is TRITONSERVER_INSTANCEGROUPKIND_AUTO right? I may have misunderstood

tanmayv25: TRITONSERVER_INSTANCEGROUPKIND_AUTO should be treated exactly like TRITONSERVER_INSTANCEGROUPKIND_GPU when adding the Execution Providers.

Assume a user providing the following config:

```
instance_group [ { kind: KIND_AUTO }]

optimization { execution_accelerators {
  gpu_execution_accelerator : [ {
    name : "tensorrt"
    parameters { key: "precision_mode" value: "FP16" }
    parameters { key: "max_workspace_size_bytes" value: "1073741824" }
    }]
}}
``` 
We need to add the TensorRT EP in this case.

tanmayv25 (approved):
@CoderHam Did the test pass locally with this change?

tanmayv25:
@GuanLuo  Can you also take a look on these changes?

pranavsharma:
Can we update the documentation to state that auto generated config defaults to GPU provider with default settings?

pranavsharma (approved):

GuanLuo (approved):
Should make sure the backend built with TRT / CUDA EP works on system that has no GPU, i.e. model will be loaded on CPU successfully

CoderHam:
>@CoderHam Did the test pass locally with this change?

Yes they passed locally. Running exhaustive testing in CI as well


> Can we update the documentation to state that auto generated config defaults to execution on the GPU with default settings?

Sure. @pranavsharma do you want this documented inside this repo or in the server autocomplete docs?

> Should make sure the backend built with TRT / CUDA EP works on system that has no GPU, i.e. model will be loaded on CPU successfully

Testing this manually since it is not tested in our CI for the case when autofill is enabled.

pranavsharma:
> > Can we update the documentation to state that auto generated config defaults to execution on the GPU with default settings?
> 
> Sure. @pranavsharma do you want this documented inside this repo or in the server autocomplete docs?
> 
The autocomplete feature is mentioned in the server docs, so I guess we can add it there.

PR #114, opened 2022-04-22
Default to load model on GPU when instance kind is not specified  (#112)

CoderHam:
* Default to load model on GPU when instance kind is not specified and autocomplete is used

* Add compile flags

* cleanup

* review edits


CoderHam:
cherry-picking over to main branch

mc-nv (approved):

PR #115, opened 2022-04-22
Default to load model on GPU when instance kind is not specified  (#1…

mc-nv:
…12) (#114)

* Default to load model on GPU when instance kind is not specified and autocomplete is used

* Add compile flags

* cleanup

* review edits


CoderHam:
Closing since this change is already in https://github.com/triton-inference-server/onnxruntime_backend/tree/r22.04

PR #118, opened 2022-05-06
Fix Autocomplete To Use CPU When No GPU Found

nv-kmcgill53:
This PR fixes onnxruntime's autocomplete to check the instance(s) kind of the of the model before loading. If no GPU instance groups are found then autocomplete will set the kind to CPU before loading the model.

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -692,7 +692,6 @@ ModelState::AutoCompleteConfig()
     // have already occurred. If at least one instance group with
     // "kind" = "KIND_GPU" then allow model to use GPU else autocomplete to
     // "KIND_CPU"
-    bool found_gpu_instance = false;
     for (size_t i = 0; i < instance_group.ArraySize(); ++i) {
       triton::common::TritonJson::Value instance_obj;
       instance_group.IndexAsObject(i, &instance_obj);
@@ -703,15 +702,12 @@ ModelState::AutoCompleteConfig()
       RETURN_IF_ERROR(instance_group_kind.AsString(&kind_str));
 
       if (kind_str == "KIND_GPU") {
-        found_gpu_instance = true;
+        kind = TRITONSERVER_INSTANCEGROUPKIND_GPU;
         break;
       }
     }
-
-    if (found_gpu_instance) {
-      kind = TRITONSERVER_INSTANCEGROUPKIND_GPU;
-    }
 #endif  // TRITON_ENABLE_GPU
+
     OrtSession* sptr = nullptr;
     RETURN_IF_ERROR(LoadModel(
         artifact_name, kind, 0, &model_path, &sptr, &default_allocator,
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -682,14 +682,16 @@ ModelState::AutoCompleteConfig()
   OrtAllocator* default_allocator;
   std::string model_path;
   {
-    TRITONSERVER_InstanceGroupKind kind;
+    TRITONSERVER_InstanceGroupKind kind = TRITONSERVER_INSTANCEGROUPKIND_CPU;
 
+#ifdef TRITON_ENABLE_GPU
     triton::common::TritonJson::Value instance_group;
     ModelConfig().Find("instance_group", &instance_group);
 
     // Earlier in the model lifecycle, device checks for the instance group
     // have already occurred. If at least one instance group with
-    // "kind" = "KIND_GPU" then allow model to use GPU else autocomplete to CPU
+    // "kind" = "KIND_GPU" then allow model to use GPU else autocomplete to
+    // "KIND_CPU"
     bool found_gpu_instance = false;
     for (size_t i = 0; i < instance_group.ArraySize(); ++i) {
       triton::common::TritonJson::Value instance_obj;
@@ -700,7 +702,7 @@ ModelState::AutoCompleteConfig()
       std::string kind_str;
       RETURN_IF_ERROR(instance_group_kind.AsString(&kind_str));
 
-      if (kind_str == "KIND_GPU" || kind_str == "KIND_AUTO") {
+      if (kind_str == "KIND_GPU") {
         found_gpu_instance = true;
         break;
       }
@@ -708,10 +710,8 @@ ModelState::AutoCompleteConfig()
 
     if (found_gpu_instance) {
       kind = TRITONSERVER_INSTANCEGROUPKIND_GPU;
-    } else {
-      kind = TRITONSERVER_INSTANCEGROUPKIND_CPU;
     }
-
+#endif  // TRITON_ENABLE_GPU
     OrtSession* sptr = nullptr;
     RETURN_IF_ERROR(LoadModel(
         artifact_name, kind, 0, &model_path, &sptr, &default_allocator,
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -682,10 +682,40 @@ ModelState::AutoCompleteConfig()
   OrtAllocator* default_allocator;
   std::string model_path;
   {
+    TRITONSERVER_InstanceGroupKind kind;
+
+    triton::common::TritonJson::Value instance_group;
+    ModelConfig().Find("instance_group", &instance_group);
+
+    // Earlier in the model lifecycle, device checks for the instance group
+    // have already occurred. If at least one instance group with
+    // "kind" = "KIND_GPU" then allow model to use GPU else autocomplete to CPU
+    bool found_gpu_instance = false;
+    for (size_t i = 0; i < instance_group.ArraySize(); ++i) {
+      triton::common::TritonJson::Value instance_obj;
+      instance_group.IndexAsObject(i, &instance_obj);
+
+      triton::common::TritonJson::Value instance_group_kind;
+      instance_obj.Find("kind", &instance_group_kind);
+      std::string kind_str;
+      RETURN_IF_ERROR(instance_group_kind.AsString(&kind_str));
+
+      if (kind_str == "KIND_GPU" || kind_str == "KIND_AUTO") {
+        found_gpu_instance = true;
+        break;
+      }
+    }
+
+    if (found_gpu_instance) {
+      kind = TRITONSERVER_INSTANCEGROUPKIND_GPU;
+    } else {
+      kind = TRITONSERVER_INSTANCEGROUPKIND_CPU;
+    }
+
     OrtSession* sptr = nullptr;
     RETURN_IF_ERROR(LoadModel(
-        artifact_name, TRITONSERVER_INSTANCEGROUPKIND_AUTO, 0, &model_path,
-        &sptr, &default_allocator, nullptr));
+        artifact_name, kind, 0, &model_path, &sptr, &default_allocator,
+        nullptr));
     session.reset(sptr);
   }
   OnnxTensorInfoMap input_tensor_infos;


src/onnxruntime.cc
```diff
@@ -682,10 +682,40 @@ ModelState::AutoCompleteConfig()
   OrtAllocator* default_allocator;
   std::string model_path;
   {
+    TRITONSERVER_InstanceGroupKind kind;
```
CoderHam, line 685: Initialize this to CPU by default and change to GPU when 
```
      if (kind_str == "KIND_GPU" || kind_str == "KIND_AUTO") {
        kind = TRITONSERVER_INSTANCEGROUPKIND_GPU;
        break;
      }
```

src/onnxruntime.cc
```diff
@@ -682,10 +682,40 @@ ModelState::AutoCompleteConfig()
...
+    for (size_t i = 0; i < instance_group.ArraySize(); ++i) {
+      triton::common::TritonJson::Value instance_obj;
+      instance_group.IndexAsObject(i, &instance_obj);
+
+      triton::common::TritonJson::Value instance_group_kind;
+      instance_obj.Find("kind", &instance_group_kind);
+      std::string kind_str;
+      RETURN_IF_ERROR(instance_group_kind.AsString(&kind_str));
+
+      if (kind_str == "KIND_GPU" || kind_str == "KIND_AUTO") {
```
tanmayv25, line 703: Is it possible for `kind_str == "KIND_AUTO"`? I believe model_config_utils.cc in Triton core will override KIND_AUTO to the correct KIND_CPU/KIND_GPU.

src/onnxruntime.cc
```diff
@@ -682,10 +682,40 @@ ModelState::AutoCompleteConfig()
...
   std::string model_path;
   {
+    TRITONSERVER_InstanceGroupKind kind;
+
+    triton::common::TritonJson::Value instance_group;
+    ModelConfig().Find("instance_group", &instance_group);
+
+    // Earlier in the model lifecycle, device checks for the instance group
+    // have already occurred. If at least one instance group with
+    // "kind" = "KIND_GPU" then allow model to use GPU else autocomplete to CPU
```
CoderHam, line 692: For consistency change to "autocomplete to "KIND_CPU".

src/onnxruntime.cc
tanmayv25: I believe if the bakcend gets KIND_AUTO from Triton Core here, then we must raise an error instead of taking it as GPU instance.

src/onnxruntime.cc
```diff
@@ -682,10 +682,40 @@ ModelState::AutoCompleteConfig()
...
+    bool found_gpu_instance = false;
```
CoderHam, line 693: Can be removed. Refer to earlier suggestion.

src/onnxruntime.cc
```diff
@@ -682,10 +682,40 @@ ModelState::AutoCompleteConfig()
   OrtAllocator* default_allocator;
   std::string model_path;
   {
+    TRITONSERVER_InstanceGroupKind kind;
+
+    triton::common::TritonJson::Value instance_group;
```
CoderHam, line 687: From here to L713 should be wrapped inside TRITON_ENABLE_GPU. If it is a CPU only build, should always autofill kind to KIND_CPU

CoderHam (changes requested):

src/onnxruntime.cc
```diff
@@ -682,10 +682,40 @@ ModelState::AutoCompleteConfig()
...
+
+      if (kind_str == "KIND_GPU" || kind_str == "KIND_AUTO") {
+        found_gpu_instance = true;
+        break;
+      }
+    }
+
+    if (found_gpu_instance) {
+      kind = TRITONSERVER_INSTANCEGROUPKIND_GPU;
+    } else {
```
rmccorm4, line 711: 1. Isn't the problem in https://github.com/triton-inference-server/server/issues/4346#issuecomment-1119885043 because default KIND is KIND_AUTO ? Wouldn't this code still default to GPU if no kind is specified even on a CPU machine? Or can you point me to the code that sets KIND to KIND_CPU if no GPUs are available? (or include a comment detailing similarly)

2. Should this check be done in `LoadModel()` instead since it can be passed KIND_AUTO even if GPU isn't necessarily supported? Currently KIND_AUTO == KIND_GPU in LoadModel(), and it seems more like KIND_AUTO should specify that we pick GPU if available, otherwise pick CPU.

rmccorm4 (changes requested):

src/onnxruntime.cc
nv-kmcgill53: I looked through the code again (https://github.com/triton-inference-server/core/blob/2ba05e79e8f5db7c354f2ffb48cc8cd920c23be7/src/model_config_utils.cc#L749). You are right, KIND_AUTO should be processed appropriately. I am taking out the "KIND_AUTO" check.

src/onnxruntime.cc
nv-kmcgill53: 1. The previous code I was referring to was [here]((https://github.com/triton-inference-server/core/blob/2ba05e79e8f5db7c354f2ffb48cc8cd920c23be7/src/model_config_utils.cc#L749). If the model config gets loaded with KIND_AUTO, then the linked code checks for devices which it found and assigns the instance group kind. In the case of the bug this is solving, the `supported_gpus` set will be empty since no devices were found and the group instance kind will be set to KIND_CPU.  "KIND_AUTO" should never make it to the backend, which should address your concern.
2. I hesitate putting this into `LoadModel()` since `ModelInstanceState()` also [calls this function](https://github.com/triton-inference-server/onnxruntime_backend/blob/493d43ef21906df32ec80d99676f19adfc2176af/src/onnxruntime.cc#L949), and I'm not certain what code is getting executed before this. Also, both the offline channel discussion and the linked github issue both specify using `--strict-model-config=false` which is what this PR aims to solve. I can look further into this if you wish.

src/onnxruntime.cc
nv-kmcgill53: The issue our users are seeing is that the onnxruntime_backend has been compiled with 'TRITON_ENABLE_GPU' and LoadModel() is getting passed a parameter which specifies to use the GPU when there is none. Wrapping this code in an `#ifdef` doesn't solve the problem since we still have to check the instance group kind. With that said, after looking at your suggested change, I agree that the code reads better for this section as it simplifies the logic of "autocomplete will only verify the instance kind if compiled with TRITON_ENABLE_GPU." If the backend is not compiled with the TRITON_ENABLE_GPU flag on, then don't worry about it.

src/onnxruntime.cc
rmccorm4: re: (2), Keeping the logic in Autocomplete is OK for now

src/onnxruntime.cc
```diff
@@ -700,18 +702,16 @@ ModelState::AutoCompleteConfig()
       std::string kind_str;
       RETURN_IF_ERROR(instance_group_kind.AsString(&kind_str));
 
-      if (kind_str == "KIND_GPU" || kind_str == "KIND_AUTO") {
+      if (kind_str == "KIND_GPU") {
         found_gpu_instance = true;
```
rmccorm4, line 706: Does it make sense to just set the kind here and break rather than toggling the bool?
```cpp
    if (kind_str == "KIND_GPU") {
      kind = TRITONSERVER_INSTANCEGROUPKIND_GPU;
      break
    }
```

src/onnxruntime.cc
nv-kmcgill53: We still have to search through the config's instances since the root of the issue is the shared lib being compiled with TRITON_ENABLE_GPU but no device was found during runtime.

src/onnxruntime.cc
nv-kmcgill53: fixed

src/onnxruntime.cc
nv-kmcgill53: Updated to `kind = TRITONSERVER_INSTANCEGROUPKIND_CPU` for initialization

src/onnxruntime.cc
nv-kmcgill53: good find!

nv-kmcgill53: Fixed

src/onnxruntime.cc
rmccorm4: re: (1) in your testing, can you confirm that loading a model config that **doesn't** specify instance_group with `--strict-model-config=true` will correctly use CPU if no GPU is available?

You should be able to use our example [densenet_onnx](https://github.com/triton-inference-server/server/blob/main/docs/examples/model_repository/densenet_onnx/config.pbtxt) model and config if you don't have any others handy.

nv-kmcgill53: Will check this out

src/onnxruntime.cc
tanmayv25: I believe the ask here was to remove `found_gpu_instance` bool variable and directly set the kind to GPU here.

src/onnxruntime.cc
nv-kmcgill53: Tested this out with the "resnet50" model (not the cpu verison) and used the following command: `$tritonserver --model-repository=/path/to/models --strict-model-config=true --log-verbose=1` and following occurred:
1. Model was successfully loaded 
2. When queried, the model configuration has the following:
  a.   ```  "instance_group": [{
            "name": "resnet50",
            "kind": "KIND_CPU",
            "count": 1,
            "gpus": [],
            "secondary_devices": [],
            "profile": [],
            "passive": false,
            "host_policy": ""
        }
    ],```

When adding the following to the config file:
```
instance_group [
    {
      count: 1
      kind: KIND_GPU
      gpus: [ 0 ]
    }
]
```
and then executing `$tritonserver --model-repository=/path/to/models --strict-model-config=true --log-verbose=1` I received the following:
1. The model did not load
2. Triton returned the expected error: `resnet50 specifies invalid or unsupported gpu id 0. GPUs with at least the minimum required CUDA compute compatibility of 6.000000`

src/onnxruntime.cc
nv-kmcgill53: Forgot to push the change after testing. updated

src/onnxruntime.cc
nv-kmcgill53: Sorry I read this comment wrong the first time. You and [rmccorm4](https://github.com/rmccorm4) had the same suggestion.

tanmayv25 (approved):

src/onnxruntime.cc
rmccorm4: Awesome, thanks! I assume both an empty list or omitting the `gpus` field here would have similar behavior?

nv-kmcgill53: 1. Omitting the GPUs field with `--strict-model-config=true` constructed the correct model config with kind=KIND_CPU and model was loaded successfully.
2. Empty list had the same effect as (1)

GuanLuo:
@pranavsharma FYI, this is follow-up fix for https://github.com/triton-inference-server/onnxruntime_backend/pull/112. Before this fix the auto-complete will set all EPs as the model may contain EP-specific ops, but there is an edge case where model will fail to load if CUDA EP is set while there is no GPU on the system. With this change, the auto-complete will check GPU availability and set EPs conditionally

rmccorm4 (approved):

CoderHam (approved):

PR #120, opened 2022-05-17
Autocomplete now makes use of the default-max-batch-size

nv-kmcgill53:
Onnxruntime backend now makes use of the default max-batch-size value which is passed to it. If the final max-batch-size is greater than 1 then the backend turns on dynamic batching if no other schedulers are found.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,8 +185,29 @@ parameters { key: "inter_op_thread_count" value: { string_value: "0" } }
 * `memory.enable_memory_arena_shrinkage`: See [this](https://github.com/microsoft/onnxruntime/blob/master/include/onnxruntime/core/session/onnxruntime_run_options_config_keys.h) for more information.
 
 ### Command line options
+
+#### Thread Pools
+
 When intra and inter op threads is set to 0 or a value higher than 1, by default ORT creates threadpool per session. This may not be ideal in every scenario, therefore ORT also supports global threadpools. When global threadpools are enabled ORT creates 1 global threadpool which is shared by every session. Use the backend config to enable global threadpool. When global threadpool is enabled, intra and inter op num threads config should also be provided via backend config. Config values provided in model config will be ignored.
 
 ```
 --backend-config=onnxruntime,enable-global-threadpool=<0,1>, --backend-config=onnxruntime,intra_op_thread_count=<int> , --backend-config=onnxruntime,inter_op_thread_count=<int> 
 ```
+
+#### Default Max Batch Size
+
+The default-max-batch-size value is used for max_batch_size during [Autocomplete](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#auto-generated-model-configuration) when no 
+other value is found. If the `--strict-model-config=false` command-line
+option is used, the onnxruntime backend will set the max_batch_size
+of the model to this default value under the following conditions:
+
+1. Autocomplete has determined the model is capable of batching requests. 
+2. max_batch_size is 0 in the model configuration or max_batch_size 
+   is omitted from the model configuration.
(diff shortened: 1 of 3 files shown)


nv-kmcgill53:
Added section in README explaining the default-max-batch-size configuration.

README.md
```diff
@@ -185,8 +185,29 @@ parameters { key: "inter_op_thread_count" value: { string_value: "0" } }
...
+
 When intra and inter op threads is set to 0 or a value higher than 1, by default ORT creates threadpool per session. This may not be ideal in every scenario, therefore ORT also supports global threadpools. When global threadpools are enabled ORT creates 1 global threadpool which is shared by every session. Use the backend config to enable global threadpool. When global threadpool is enabled, intra and inter op num threads config should also be provided via backend config. Config values provided in model config will be ignored.
 
 ```
 --backend-config=onnxruntime,enable-global-threadpool=<0,1>, --backend-config=onnxruntime,intra_op_thread_count=<int> , --backend-config=onnxruntime,inter_op_thread_count=<int> 
 ```
+
+#### Default Max Batch Size
+
+The default-max-batch-size value is used for max_batch_size during [Autocomplete](https://github.com/triton-inference-server/server/blob/main/docs/model_configuration.md#auto-generated-model-configuration) when no
```
tanmayv25, line 199: This is fine for now. But in future we will have a separate new section in this README for describing how it works with the auto-complete feature.
We can then link to this backend-specific section here.

src/onnxruntime.cc
```diff
@@ -2484,6 +2521,22 @@ TRITONBACKEND_Initialize(TRITONBACKEND_Backend* backend)
   // Onetime initialization for the onnxruntime loader.
   RETURN_IF_ERROR(OnnxLoader::Init(backend_config));
 
+  std::unique_ptr<BackendConfiguration> lconfig(new BackendConfiguration());
```
tanmayv25, line 2524: We are already reading the backend_config in `OnnxLoader::Init` for the thread pool config. Does it make sense to parse the backend_config once and pass around the required information to these functions?

nv-kmcgill53: I thought about that and came to the conclusion it might be easier to maintain if all backends had the same look to them.  I'll go ahead and change this.

nv-kmcgill53: After the attempted change, I found that the code layout would look worse than it does right now. To make the suggested change, the `BackendConfiguration` struct would have to move into onnxruntime_loader.h or I would have to create a new header file onnxruntime.h which only holds the definition of the struct.  I'm not a fan of both of these options since it diverges the general backend code base without a good reason for it, imo. I can make an argument for processing onnx configs in the Init() function and parsing general backned parameters in onnxruntime.cc but let me know.

tanmayv25: We can pass only the required parameters from backend config to OnnxLoader::Init().

tanmayv25: Resolving the comment for now to limit the scope of the change.

tanmayv25 (approved):

src/onnxruntime.cc
```diff
@@ -765,21 +772,55 @@ ModelState::AutoCompleteMaxBatch(
   // supported and max-batch-size is not specified. We need to update
   // the configuration itself as well as the cached value we have already
   // initialized in the model state.
+#define KYLE_DEFINE 3
```
GuanLuo, line 775: Remove this define?

nv-kmcgill53: Took out debug define

tanmayv25 (approved):

PR #126, opened 2022-07-06
Fixes tensorrt cache being regenerated on path change

fran6co:
Onnxruntime uses the path as a hash for the tensorrt cache, passing a binary avoids this and works as intended.

Fixes https://github.com/triton-inference-server/server/issues/4587

### DIFF
--- a/src/onnxruntime_loader.cc
+++ b/src/onnxruntime_loader.cc
@@ -31,6 +31,7 @@
 #include <locale>
 #include <string>
 #include <thread>
+#include <fstream>
 
 #include "onnxruntime_utils.h"
 
@@ -191,8 +192,13 @@ OnnxLoader::LoadSession(
           loader->env_, ort_style_model_str.c_str(), model.size(),
           session_options, session);
     } else {
-      status = ort_api->CreateSession(
-          loader->env_, ort_style_model_str.c_str(), session_options, session);
+      std::ifstream bytes_stream(ort_style_model_str.c_str(), std::ios::in | std::ios::binary | std::ios::ate);
+      std::streamsize num_bytes = bytes_stream.tellg();
+      bytes_stream.seekg(0, std::ios::beg);
+      std::vector<char> buffer(num_bytes);
+      bytes_stream.read(buffer.data(), num_bytes);
+      status = ort_api->CreateSessionFromArray(
+          loader->env_, buffer.data(), num_bytes, session_options, session);
     }
 
     if (status != nullptr) {
--- a/src/onnxruntime_loader.cc
+++ b/src/onnxruntime_loader.cc
@@ -192,6 +192,10 @@ OnnxLoader::LoadSession(
           loader->env_, ort_style_model_str.c_str(), model.size(),
           session_options, session);
     } else {
+      // Onnxruntime uses the model path to create the TRT hash path,
+      // this triggers a regeneration when coupled with temporal paths
+      // Using the binary directly sidesteps the issue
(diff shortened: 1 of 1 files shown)


GuanLuo:
@pranavsharma can you or someone familiar with TRT EP take a look at this PR? The change looks fine to me and just want to double check if the cache behavior will be changed as intended.

pranavsharma:
@stevenlix - any comments? Is this how we avoid TRT cache regeneration?

stevenlix:
Yes. ORT-TRT uses ModelMetadefIdGenerator function to generate engine id. If the model passed in has a path, the id will be generated from the path string (https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/core/framework/execution_provider.cc#L151). To avoid the dependency on path, one can pass model binary to the inference session.

cnegron-nv:
@pranavsharma Can you give a final review to this PR?

pranavsharma (dismissed):

tanmayv25 (approved):

src/onnxruntime_loader.cc
```diff
@@ -191,8 +192,13 @@ OnnxLoader::LoadSession(
           loader->env_, ort_style_model_str.c_str(), model.size(),
           session_options, session);
     } else {
-      status = ort_api->CreateSession(
-          loader->env_, ort_style_model_str.c_str(), session_options, session);
+      std::ifstream bytes_stream(ort_style_model_str.c_str(), std::ios::in | std::ios::binary | std::ios::ate);
```
Tabrizian, line 195: Can you add a comment mentioning that the reason to pass the ONNX file as a binary is to avoid the TRT cache from being regenerated?

Tabrizian (approved):

rmccorm4 (approved):

Tabrizian:
Hi @fran6co, @pranavsharma, @stevenlix 

We are observing a failure in the CI as a result of this change. Looks like for models that expect additional weight files the relative paths do not work properly anymore. The test fails with the error below:

```
Deserialize tensor embeddings.20.weight failed.open file "./embeddings.20.weight" failed: No such file or directory |
```

Do you have a fix in mind for this? Otherwise, I think we need to revert this change.

pranavsharma:
I've reached out to @stevenlix. When is the latest you want this fix? If it's blocking the release feel free to revert it.

Tabrizian:
Thanks for the quick response! We are code freezing by the end of next week but I think it would be great if it is possible to have the fix ready for review by Wednesday next so that we can test it before we are code frozen.

pranavsharma:
> Hi @fran6co, @pranavsharma, @stevenlix
> 
> We are observing a failure in the CI as a result of this change. Looks like for models that expect additional weight files the relative paths do not work properly anymore. The test fails with the error below:
> 
> ```
> Deserialize tensor embeddings.20.weight failed.open file "./embeddings.20.weight" failed: No such file or directory |
> ```
> 
> Do you have a fix in mind for this? Otherwise, I think we need to revert this change.

fyi @jywu-msft

stevenlix:
@skottmckay, one solution is to not use absolute model path in https://github.com/microsoft/onnxruntime/blob/master/onnxruntime/core/framework/execution_provider.cc#L151

pranavsharma:
Let's revert this at this point because using byte stream won't work with external weight files. So, this PR *will* fail.

bamdadd:
Hi, @Tabrizian Do you plan to revert the revert anytime soon?

Tabrizian:
The underlying issue has not been fixed yet so I don't think there is any plan to revert this change.

CC @pranavsharma

PR #139, opened 2022-08-09
Migrate OV to 2022.1

kthui:


CMakeLists.txt
```diff
@@ -158,8 +158,8 @@ else()
     set(ONNXRUNTIME_LIBRARY "libonnxruntime.so")
   endif() # WIN32
   if(${TRITON_ENABLE_ONNXRUNTIME_OPENVINO})
-    set(TBB_LIBRARY "libtbb.so")
-    set(INFERENCE_ENGINE_LIBRARY "libinference_engine.so")
+    set(TBB_LIBRARY "libgna.so")
```
tanmayv25, line 161: TBB and GNA are two different things right?
Does this mean we no longer need to link against ;ibtbb.so? But we need to link with libgna.so instead?

If this is the case, then I would request to rename TBB_LIBRARY to GNA_LIBRARY. And replace all the occurence of TBB with GNA.

CMakeLists.txt
```diff
@@ -158,8 +158,8 @@ else()
...
+    set(INFERENCE_ENGINE_LIBRARY "libopenvino.so")
```
tanmayv25, line 162: INFERENCE_ENGINE_LIBRARY -> OV_LIBRARY.

tools/gen_ort_dockerfile.py
```diff
@@ -121,21 +121,22 @@ def dockerfile_for_linux(output_file):
...
 ARG ONNXRUNTIME_OPENVINO_VERSION
-ENV INTEL_OPENVINO_DIR /opt/intel/openvino_${ONNXRUNTIME_OPENVINO_VERSION}
-ENV LD_LIBRARY_PATH $INTEL_OPENVINO_DIR/deployment_tools/inference_engine/lib/intel64:$INTEL_OPENVINO_DIR/deployment_tools/ngraph/lib:$INTEL_OPENVINO_DIR/deployment_tools/inference_engine/external/tbb/lib:/usr/local/openblas/lib:$LD_LIBRARY_PATH
+ENV INTEL_OPENVINO_DIR /opt/intel/openvino_2022
+ENV LD_LIBRARY_PATH $INTEL_OPENVINO_DIR/tools/compile_tool:$INTEL_OPENVINO_DIR/runtime/3rdparty/tbb/lib:$INTEL_OPENVINO_DIR/runtime/3rdparty/hddl/lib:$INTEL_OPENVINO_DIR/runtime/lib/intel64:/usr/local/openblas/lib:$LD_LIBRARY_PATH
 ENV PYTHONPATH $INTEL_OPENVINO_DIR/tools:$PYTHONPATH
-ENV IE_PLUGINS_PATH $INTEL_OPENVINO_DIR/deployment_tools/inference_engine/lib/intel64
-ENV InferenceEngine_DIR=$INTEL_OPENVINO_DIR/deployment_tools/inference_engine/share
-ENV ngraph_DIR=$INTEL_OPENVINO_DIR/deployment_tools/ngraph/cmake
+ENV IE_PLUGINS_PATH $INTEL_OPENVINO_DIR/runtime/lib/intel64
```
tanmayv25, line 127: I think these names here need an update. Openvino doesn't have inference_engine or ngraph abstraction anymore right?

tools/gen_ort_dockerfile.py
```diff
@@ -266,48 +267,21 @@ def dockerfile_for_linux(output_file):
...
-    cp /opt/intel/openvino_${ONNXRUNTIME_OPENVINO_VERSION}/deployment_tools/inference_engine/external/tbb/lib/libtbb.so.2 \
-       /opt/onnxruntime/lib && \
-    cp /opt/intel/openvino_${ONNXRUNTIME_OPENVINO_VERSION}/deployment_tools/inference_engine/lib/intel64/plugins.xml \
-       /opt/onnxruntime/lib && \
-    cp /opt/intel/openvino_${ONNXRUNTIME_OPENVINO_VERSION}/deployment_tools/inference_engine/lib/intel64/libMKLDNNPlugin.so \
-       /opt/onnxruntime/lib && \
-    cp /opt/intel/openvino_${ONNXRUNTIME_OPENVINO_VERSION}/deployment_tools/inference_engine/lib/intel64/libinference_engine_lp_transformations.so \
-       /opt/onnxruntime/lib && \
-    cp /opt/intel/openvino_${ONNXRUNTIME_OPENVINO_VERSION}/deployment_tools/inference_engine/lib/intel64/libinference_engine_ir_reader.so \
+    cp /opt/intel/openvino_2022/runtime/lib/intel64/lib*.so \
```
tanmayv25, line 277: Why we are copying all the libraries here? Can we only copy the subset that might be sufficient like we used to do before?
There might be some runtime libraries designed for different hardware that we don't support. I am concerned about the size and possible symbol collisions it may cause. If there aren't many unneeded libraries, I think this way of copying all the libraries future-proofs us.

tanmayv25 (changes requested):

tools/gen_ort_dockerfile.py
kthui: Actually, openvino retained the inference_engine and ngraph, and simply added the openvino api 2.0 with the 2022 release. The new `ov` namespace also have the old `InferenceEngine` on top of it [here](https://github.com/openvinotoolkit/openvino/blob/2022.1.0/src/core/include/openvino/core/any.hpp#L23).
The paths are needed for a successful build.

CMakeLists.txt
kthui: My mistake. Should be `libtbb.so`. Updated.

CMakeLists.txt
kthui: Updated.

tools/gen_ort_dockerfile.py
kthui: Yes. The list is trimmed down.

tanmayv25 (approved):

PR #150, opened 2022-10-21
Add ORT 1.13.1 support

mc-nv:


tools/gen_ort_dockerfile.py
```diff
@@ -165,7 +165,7 @@ def dockerfile_for_linux(output_file):
 
        '''
     # Use the tensorrt-8.5ea branch to use Tensor RT 8.5a to use the built-in tensorrt parser
-    elif FLAGS.ort_version == "1.12.1" or FLAGS.ort_version == "1.13.0":
+    elif FLAGS.ort_version == "1.12.1" or FLAGS.ort_version == "1.13.0" or  FLAGS.ort_version == "1.13.1" :
```
dzier, line 168: Why do we reference 1.12.1 here?

dzier (approved):

nv-kmcgill53 (approved):

tools/gen_ort_dockerfile.py
nv-kmcgill53: 1.12.1 was ORT upgrading to use the EA version of TRT 8.5 for 22.08

PR #151, opened 2022-10-24
Fixing condition for ORT 1.13.1

mc-nv:


tools/gen_ort_dockerfile.py
```diff
@@ -208,7 +208,7 @@ def dockerfile_for_linux(output_file):
             ep_flags += ' --cudnn_home "{}"'.format(FLAGS.cudnn_home)
         if FLAGS.ort_tensorrt:
             ep_flags += ' --use_tensorrt'
-            if FLAGS.ort_version == "1.12.1" or FLAGS.ort_version == "1.13.0" :
+            if FLAGS.ort_version == "1.12.1" or FLAGS.ort_version == "1.13.0" or FLAGS.ort_version == "1.13.1" :
```
tanmayv25, line 211: @mc-nv  Could you try with:
```
if FLAGS.ort_version == "1.12.1":
   ep_flags += ' --use_tensorrt_builtin_parser'
```
I think built in parser is only needed for ort 1.12.1. @pranavsharma Can you confirm? Can ORT `rel-1.13.1` be built as it is?

tanmayv25:
@pranavsharma  can you review these changes?

tools/gen_ort_dockerfile.py
pranavsharma: cc @jywt-msft I don't think we need this condition for 1.13.1.

tanmayv25 (approved):

PR #174, opened 2023-03-11
Serialize ModelInstanceState constructor for parallel model loading

kthui:
Related PR: https://github.com/triton-inference-server/core/pull/165

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -1005,13 +1005,6 @@ ModelInstanceState::ModelInstanceState(
       cuda_allocator_info_(nullptr), cpu_allocator_info_(nullptr),
       io_binding_(nullptr), output_buffer_(nullptr)
 {
-  // Segmentation fault on server destruction is observed under parallel model
-  // loading when running L0_onnx_optimization if this construction is not
-  // serialized.
-  // [FIXME] TITLE (DLIS-XXXX)
-  static std::mutex construction_mu;
-  std::lock_guard<std::mutex> construction_lock(construction_mu);
-
   THROW_IF_BACKEND_INSTANCE_ERROR(model_state->LoadModel(
       ArtifactFilename(), Kind(), DeviceId(), &model_path_, &session_,
       &default_allocator_, CudaStream()));
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -1005,6 +1005,13 @@ ModelInstanceState::ModelInstanceState(
       cuda_allocator_info_(nullptr), cpu_allocator_info_(nullptr),
       io_binding_(nullptr), output_buffer_(nullptr)
 {
+  // Segmentation fault on server destruction is observed under parallel model
+  // loading when running L0_onnx_optimization if this construction is not
+  // serialized.
+  // [FIXME] TITLE (DLIS-XXXX)
+  static std::mutex construction_mu;
+  std::lock_guard<std::mutex> construction_lock(construction_mu);
+
   THROW_IF_BACKEND_INSTANCE_ERROR(model_state->LoadModel(
       ArtifactFilename(), Kind(), DeviceId(), &model_path_, &session_,
       &default_allocator_, CudaStream()));
--- a/src/onnxruntime_loader.cc
+++ b/src/onnxruntime_loader.cc
@@ -186,13 +186,19 @@ OnnxLoader::LoadSession(
     }
 
     OrtStatus* status = nullptr;
-    if (!is_path) {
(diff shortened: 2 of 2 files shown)


src/onnxruntime.cc
```diff
@@ -1005,6 +1005,13 @@ ModelInstanceState::ModelInstanceState(
       cuda_allocator_info_(nullptr), cpu_allocator_info_(nullptr),
       io_binding_(nullptr), output_buffer_(nullptr)
 {
+  // Segmentation fault on server destruction is observed under parallel model
+  // loading when running L0_onnx_optimization if this construction is not
+  // serialized.
+  // [FIXME] TITLE (DLIS-XXXX)
+  static std::mutex construction_mu;
+  std::lock_guard<std::mutex> construction_lock(construction_mu);
```
GuanLuo, line 1013: Is this related to appending EP on to model loading resulting in EP initialization which must ensure exclusive access? If so, can we take a more finer control to only provide exclusive access when the model loading involves EP? CC @pranavsharma

src/onnxruntime.cc
```diff
@@ -1005,6 +1005,13 @@ ModelInstanceState::ModelInstanceState(
       cuda_allocator_info_(nullptr), cpu_allocator_info_(nullptr),
       io_binding_(nullptr), output_buffer_(nullptr)
 {
+  // Segmentation fault on server destruction is observed under parallel model
+  // loading when running L0_onnx_optimization if this construction is not
+  // serialized.
+  // [FIXME] TITLE (DLIS-XXXX)
```
dyastremsky, line 1011: Don't forget to add this information (needed before approval).

kthui: Yes, updated along with the commit: [Reduce lock scope to ort create session](https://github.com/triton-inference-server/onnxruntime_backend/pull/174/commits/2899f7c90f817a41470967f2fe41aa1fc9fc4d56)

src/onnxruntime.cc
kthui: > Is this related to appending EP on to model loading resulting in EP initialization which must ensure exclusive access?

Nice suggestion! I was able to reduce the lock scope down to only serializing the ORT create session API. The segfault was not observed after ~128 consecutive loops, while the segfault was observed after ~6 loops without adding the lock. 
Update commit: [Reduce lock scope to ort create session](https://github.com/triton-inference-server/onnxruntime_backend/pull/174/commits/2899f7c90f817a41470967f2fe41aa1fc9fc4d56)

kthui: @pranavsharma FYI, segfault on destruction was observed if the create session API is called concurrently:
```
Thread 9 "tritonserver" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7fe7d0ffd000 (LWP 4225)]
0x00007fe7d8424230 in ?? ()
(gdb) thread apply all bt
...
Thread 9 (Thread 0x7fe7d0ffd000 (LWP 4225)):
#0  0x00007fe7d8424230 in ?? ()
#1  0x00007fe7e5f305a1 in __nptl_deallocate_tsd () at pthread_create.c:301
#2  0x00007fe7e5f3162a in __nptl_deallocate_tsd () at pthread_create.c:256
#3  start_thread (arg=<optimized out>) at pthread_create.c:488
#4  0x00007fe7e48a5133 in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:95
...
Thread 1 (Thread 0x7fe7e354d000 (LWP 4211)):
#0  __pthread_clockjoin_ex (threadid=140633620598784, thread_return=0x0, clockid=<optimized out>, abstime=<optimized out>, block=<optimized out>) at pthread_join_common.c:145
#1  0x00007fe7e4a4f047 in std::thread::join() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#2  0x00007fe7e50f6f54 in triton::common::ThreadPool::~ThreadPool (this=0x5617cfa281b0, __in_chrg=<optimized out>) at /tmp/tritonbuild/tritonserver/build/_deps/repo-core-build/triton-core/_deps/repo-common-src/src/thread_pool.cc:80
...
#20 main (argc=<optimized out>, argv=<optimized out>) at /opt/server/src/main.cc:459
```

Tabrizian (approved):

tanmayv25 (approved):

src/onnxruntime.cc
GuanLuo: Segfault with **AND** without execution provider, if exclusive access not ensured?

GuanLuo (approved):

PR #177, opened 2023-03-22
Support initializer as inference request input - onnxruntime backend

taoisu:
Background of this PR can be found here: https://github.com/triton-inference-server/server/issues/5489

Essentially, in onnxruntime, initializers (model weights) can be specified as input to override the default weights. The initializers are different in the sense they don't have batch dimension, so changes need to be made in this repo to make sure the shape treatment is correct. In addition, because model weights will be overridden, dynamic batching should be disabled when initializers are to be fed.

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -1315,6 +1315,15 @@ ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
     bool io_optional;
     RETURN_IF_ERROR(io.MemberAsBool("optional", &io_optional));
 
+    if (io_optional && model_state_->MaxBatchSize() != 0) {
+      return TRITONSERVER_ErrorNew(
+          TRITONSERVER_ERROR_INVALID_ARG,
+          (std::string("unable to load model '") + model_state_->Name() +
+           "', optional input '" + io_name +
+           "' is not supported for models that support batching")
+              .c_str());
+    }
+
     const auto& tensor_names = io_optional
       ? overridable_initializer_tensor_names
       : input_tensor_names;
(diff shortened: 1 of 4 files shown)


src/onnxruntime.cc
```diff
@@ -1817,7 +1856,7 @@ ModelInstanceState::SetInputTensors(
     else {
       batchn_shape =
           std::vector<int64_t>(input_shape, input_shape + input_dims_count);
-      if (max_batch_size != 0) {
+      if (max_batch_size != 0 && !is_initializer_tensor) {
```
GuanLuo, line 1859: You probably don't need to condition on this explicitly, if the model config is successfully validated, then the model shouldn't allow dynamic batching since it is the current limitation of exposing overridable initializer as input. You may enforce it explicitly in validate function that when an input is marked optional, `model_state_->MaxBatchSize()` must be 0.

GuanLuo: Then the new API and surrounding work in the core repo may be removed.

taoisu:
Hi @GuanLuo, thanks for the feedback! I have resolved all of them, I'll drop the PR in the core rpo as you suggested. Please let me know if there are any other questions.

GuanLuo (approved):
LGTM, do you have an example model / test case that will use this feature? I can create a CI test out of it to validate the change. CC @pranavsharma on the use of ORT API.

GuanLuo:
If you haven't already, do you mind to follow the contribution page [here](https://github.com/triton-inference-server/server/blob/main/CONTRIBUTING.md#contributor-license-agreement-cla) to submit a signed CLA?

taoisu:
Thanks @GuanLuo !

I have a sample model you can use to add to test: [dummy.zip](https://github.com/triton-inference-server/onnxruntime_backend/files/11081970/dummy.zip)

#### Model creation code 
```python
import onnx
import torch

import numpy as np
import torch.nn as nn


class SimpleNet(nn.Module):

    def __init__(self):
        super(SimpleNet, self).__init__()
        self.fixed = nn.Parameter(torch.arange(4, dtype=torch.float32))
        self.var = nn.Parameter(torch.rand(4, dtype=torch.float32))

    def forward(self, x: torch.Tensor):
        x = x + self.fixed + self.var
        return x

model_name = 'dummy.onnx'

torch.onnx.export(
    SimpleNet(),
    torch.randn(1, 4),
    model_name,
    input_names=['in'],
    output_names=['out'],
    dynamic_axes={'in': {0: 'bs'}, 'out': {0: 'bs'}},
)

model = onnx.load(model_name)

for initializer in model.graph.initializer:
    if initializer.name == 'var':
        model.graph.input.append(onnx.helper.make_tensor_value_info(initializer.name, initializer.data_type, initializer.dims))

onnx.save(model, model_name)

import onnxruntime as ort

sess = ort.InferenceSession('dummy.onnx', providers=['CPUExecutionProvider'])

in_tsr = np.random.randn(2, 4).astype(np.float32)
print(in_tsr)

var_tsr = np.arange(4).astype(np.float32)
print(var_tsr)

print(sess.run(None, {'in': in_tsr, 'var': var_tsr}))
```

#### Triton model config
```
name: "init_ort"
platform: "onnxruntime_onnx"
max_batch_size: 0
default_model_filename: "dummy.onnx"

input [
  {
    name: "in"
    data_type: TYPE_FP32
    dims: [-1, 4]
  },
  {
    name: "var"
    data_type: TYPE_FP32
    dims: [4]
    optional: true
  }
]
output [
  {
    name: "out"
    data_type: TYPE_FP32
    dims: [ -1, 4]
  }
]

optimization {
  graph : {
    level : 1
  }
}

instance_group [
  {
    count: 1
    kind: KIND_GPU
    gpus: [ 0 ]
  }
]

parameters { key: "execution_mode" value: { string_value: "1" } }
parameters { key: "cudnn_conv_algo_search" value: { string_value: "0" } }
```

#### Triton test code
```python
import fire

import numpy as np
import tritonclient.http as httpclient


def main():
    triton_client = httpclient.InferenceServerClient(
        url='172.17.0.1:8000',
        verbose=False,
        insecure=True)

    inputs = []

    in_tsr = np.random.randn(2, 4).astype(np.float32)
    inputs.append(httpclient.InferInput('in', in_tsr.shape, 'FP32'))
    inputs[-1].set_data_from_numpy(in_tsr, binary_data=False)
    print(f'in: {in_tsr}')

    var_tsr = np.arange(4).astype(np.float32)
    inputs.append(httpclient.InferInput('var', var_tsr.shape, 'FP32'))
    inputs[-1].set_data_from_numpy(var_tsr, binary_data=False)
    print(f'var: {var_tsr}')

    outputs = []
    outputs.append(httpclient.InferRequestedOutput('out', binary_data=False))

    try:
        results = triton_client.infer(
            'init_ort',
            inputs,
            outputs=outputs,
        )
        actual_out = results.as_numpy('out')
        print(f'out: {actual_out}')
        expect_out = in_tsr + var_tsr + np.arange(4)
        all_close = np.allclose(actual_out, expect_out)
        print(f'actual_out == expect_out: {all_close}')
    except Exception as e:
        print(e)
        return
    print('Done')


if __name__ == '__main__':
    fire.Fire(main)
```

#### Test output
```
in: [[ 1.7549193  -0.9423713   0.3669717   0.20170133]
 [-0.26022    -0.05359299  0.26156545  0.7463688 ]]
var: [0. 1. 2. 3.]
out: [[ 1.7549193  1.0576286  4.366972   6.201701 ]
 [-0.26022    1.9464071  4.261565   6.746369 ]]
actual_out == expect_out: True
Done
```

Please let me know if you need anything else. Thanks!

taoisu:
Regarding the CLA, let me double check with my colleagues and circle back with an answer.

taoisu:
Hi @GuanLuo, just had a discussion with our legal colleagues, they have some concerns about the scope of the CLA, I asked them if they think it's ok that the PR is made by an NVIDIA engineer, and they told me it would be preferred that way. Do you think it's OK that you create a PR on this matter and I drop mine? I hope we can merge this PR sooner.

GuanLuo:
Okay, I will port the changes over after I tested the change, I am working on adding test case based on the model generation script you provided

GuanLuo:
@taoisu The mirrored PR with minor fixes has been merged, feel free to close this PR. https://github.com/triton-inference-server/onnxruntime_backend/pull/180

taoisu:
This is awesome! Thanks @GuanLuo!

PR #179, opened 2023-03-31
Add flags to allow platform-specific build

dyastremsky:
This pull request adds a platform build flag to the backend build. This supercedes the auto-detection of the platform, so the user can be specific with which platform they want to build on (Ubuntu, Windows, Jetpack).

This is necessary for Jetpack, because it doesn't currently support CUDA compute_90.

I also ran the formatter on these files, so there are some formatting updates.

TensorFlow PR: https://github.com/triton-inference-server/tensorflow_backend/pull/89


tools/gen_ort_dockerfile.py
```diff
@@ -218,10 +221,16 @@ def dockerfile_for_linux(output_file):
     if FLAGS.ort_openvino is not None:
         ep_flags += ' --use_openvino CPU_FP32'
 
+    # DLIS-4658: Once Jetson build supports high enough level of CUDA, include compute_90 for Jetson.
```
nv-kmcgill53, line 224: nit: Include the min version of CUDA in the comment: 11.8

tanmayv25 (dismissed):

CMakeLists.txt
```diff
@@ -94,6 +97,7 @@ set(TRITON_BUILD_ONNXRUNTIME_OPENVINO_VERSION "" CACHE STRING "ONNXRuntime OpenV
 set(TRITON_BUILD_CUDA_VERSION "" CACHE STRING "Version of CUDA install")
 set(TRITON_BUILD_CUDA_HOME "" CACHE PATH "Path to CUDA install")
 set(TRITON_BUILD_CUDNN_HOME "" CACHE PATH "Path to CUDNN install")
+set(TRITON_BUILD_PLATFORM "" CACHE STRING "Platform of build")
```
Tabrizian, line 100: I was wondering why do we need to explicitly specify this flag? What would `platform.system().lower()` return on jetson? It seems we were able to handle Windows and x86 without this flag.

dyastremsky: If you run `platform.system().lower()` on Jetson (bare metal or Docker), it'll return "linux." That's insufficient for knowing whether we're running on Jetson or not.

dyastremsky:
Requesting re-approvals due to rebasing.

CMakeLists.txt
Tabrizian: How about using a combination of `platform.system()` and `platform.machine()`?

dyastremsky: That would work. Great find. Are you okay with me creating a small Jira ticket to refactor these few lines when I get back from vacation? I'm trying to get the move to Docker out within the next few hours, so there's not enough time for another CI run with the limited number of Jetson runners.

Tabrizian: Sounds good. I'll resolve this thread.

dyastremsky: Thanks, Iman! Created ticket here: https://jirasw.nvidia.com/browse/DLIS-4883

nv-kmcgill53 (approved):

nv-kmcgill53 (approved):

PR #189, opened 2023-05-18
Cherry pick multi-GPU fix to ORT 1.15.0

GuanLuo:
@mc-nv @nv-kmcgill53 @nnshah1 should only merge this into 23.05 if we decide not to wait for ORT's cherry pick


mc-nv:
@GuanLuo  suppose we can close it

Issue #149, opened 2022-10-21
Add ONNXRuntime extensions support


jplu:
Hi,

I was wondering if you planned at some point to support the ONNXRuntime extensions detailed on their repo https://github.com/microsoft/onnxruntime-extensions

This will allows/unlock a lot of possibilities such as post or pre-processing directly from Triton model serving when using ONNX models.

pranavsharma:
How do you plan to use ORT extensions in Triton? What's your use case? cc @wenbingl

jplu:
For my own case, I would love to be able to have my data preprocessing directly available in Triton, such as tokenization. I can do it locally if I use ORT extensions and it is very convenient as it allows me to do not stick to Python.

wenbingl:
> For my own case, I would love to be able to have my data preprocessing directly available in Triton, such as tokenization. I can do it locally if I use ORT extensions and it is very convenient as it allows me to do not stick to Python.

onnxruntime-extensions can be built as either a static library or a shared library, and the former requires onnxruntime building from source. Can you share what your preference is?

jplu:
For me to have a shared library would be more flexible in order to let the users to bring their own ORT extensions.

Jul1aK0wal1k:
@jplu as far as I can tell, the code for loading in custom op libraries is already there https://github.com/triton-inference-server/onnxruntime_backend/blob/main/src/onnxruntime.cc#L642
It seems like you need to provide the `op_library_filename` in the config.pbtxt pointing to the shared libraries in order for the backend to load them

jplu:
@Jul1aK0wal1k Indeed, the code seems to be here. It would be nice to have an example for how to properly use it with [onnxruntime-extensions](https://github.com/microsoft/onnxruntime-extensions)?

Thanks.

msyulia:
@jplu 
You'll need to build the onnxruntime-extensions and place it somewhere in your Triton Inference Server docker image
Example dockerfile (assuming you have a x86/x64 CPU):
```Dockerfile

ARG BASE_IMAGE=nvcr.io/nvidia/tritonserver

FROM ${BASE_IMAGE}

WORKDIR /opt

ARG CMAKE_BINARY_URL=https://github.com/Kitware/CMake/releases/download/v3.27.0-rc2/cmake-3.27.0-rc2-linux-x86_64.sh
ARG ORTEXTENSIONS_REPO=https://github.com/microsoft/onnxruntime-extensions
[...]

PR #196, opened 2023-06-22
Update README.md with custom ops usage example

msyulia:
Related to #149 

The ability to load a custom op libary was available since at least r22.04 but was no where documented.


dyastremsky:
Actually, this is already documented here: https://github.com/triton-inference-server/server/blob/main/docs/user_guide/custom_operations.md

I know there's a lot of documentation and we're always looking at ways to make it more intuitive and easy to navigate. However, I don't think it makes sense to duplicate the documentation here if the backends supporting custom operators/libraries have them documented there.

PR #199, opened 2023-07-05
Add GitHub action to format and lint code

dyastremsky:
This pull request adds pre-commit hooks. It also includes running the pre-commit hooks on these files, plus making fixes based on them (e.g. making sure all `.py` and `.sh` executables are set to be executable and have shebangs, spelling corrections, etc.).

This pull request adds pre-commit  hooks to run the following:
- Black for formatting Python
- Flake8 for linting Python for conformance to PEP8, PyFlakes, and circular complexity (max-line-length set to match Black's 88 default/recommended character limit; this and other settings from Black's documentation [here](https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length) to avoid enforcement of limit when unnecessary).
- isort for deterministically sorting and organizing Python imports
- Clang for formatting C++, C, Java, TextProto, Proto, and CUDA
- Codespell for finding common spelling mistakes 

In addition, this adds these native pre-commit hooks, details [here](https://pre-commit.com/hooks.html):
- check-case-conflict
- check-executables-have-shebangs
- check-merge-conflict
- check-json
- check-toml
- check-yaml
- check-shebang-scripts-are-executable
- end-of-file-fixer
- mixed-line-ending
- requirements-txt-fixer
- trailing-whitespace

To run these locally, you can go into the repo and run `pip install pre-commit` and `pre-commit install`. After that has been done once, you just need to run `pre-commit run --all-files` any time you want to apply them. Once installed, it should run automatically on commit for files included in a commit. To correct spelling errors found by codespell when possible, you can install codespell (`pip install codespell` and potentially `pip install --upgrade codespell` to use the toml file in the directory) and call `codespell -w <path>`.

This pull request also adds a GitHub action so that these are run for every pull request and push to main. The core changes are in pre-commit.yml, pre-commit-config.yaml, and pyproject.toml. The rest of the changes are the effect of applying the action to the current repo.


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,41 @@
...
+# OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+# (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+# OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+
+name: pre-commit
+
+on:
+  pull_request:
+  push:
+    branches: [main]
```
nv-kmcgill53, line 32: Is this redundant since the `main` branches are protected and that any pushes onto `main` are necessarily done by pull requests? I might advocate for triggering on pushes to release branches such as `rXX.YY` since these are not protected.

test/initializer_as_input/test.sh
nv-kmcgill53, line 1: Can this file be deleted?

dyastremsky: CC: @GuanLuo

.github/workflows/pre-commit.yml
dyastremsky: Thank you for reviewing this PR. That is a really good point! I've changed this line to `branches: ['r[0-9][0-9].[0-9][0-9]']`, which should cover branches like r22.01 and r23.06.

I asked a question to the GitHub team [here](https://github.com/github/branch-deploy/issues/165). In the CodeQL template they provided, it said pull request branches need to be a subset of push branches in this configuration file, but [local testing](https://github.com/github/branch-deploy/issues/165) found that to be untrue and actually duplicate the workflow if the branch matches both patterns.

Assuming that is correct, it  could be good to update both this workflow and the CodeQL workflow to run for all PRs (or at least those to main and release branches).

Is there a reason we don't protect release branches? It looks like it can be done here for pattern matching. I'll open a conversation offline about this, because it might make sense to require PRs for both. In that case, we wouldn't need to run on merge.

dyastremsky: Based on our offline conversation, release branches have other forms of protection. As a result, I'm going to modify the pre-commit hooks to run for all pull requests. I don't see much risk there.

Based on testing, it seems like there's no reason that pull request branches need to be a subset of push branches for this workflow automation. I'm going to change this PR to run pre-commit and codeQL for all pull requests. If we merge it and see that actions keep running as expected (or I hear back from the GitHub team confirming that there's no risk), I'll update the PRs in the other repos to do the same.

nv-kmcgill53 (dismissed):

.github/workflows/codeql.yml
```diff
@@ -77,7 +71,7 @@ jobs:
     # Command-line programs to run using the OS shell.
     # See https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun
 
-    #   If the Autobuild fails above, remove it and uncomment the following three lines. 
+    #   If the Autobuild fails above, remove it and uncomment the following three lines.
```
jbkyang-nvi, line 74: why do we need to add codeql just for the onnxruntime repo? The other repos don't need this?

dyastremsky: CodeQL won't build out-of-the-box for our C++ structure (and it might have some issues with our Java/Go files, though I don't remember). So we only use it for Python. If we have no Python files in a repo, it automatically fails due to not finding any files, so we only have it in the repos that have at least one Python file.

dyastremsky:
@jbkyang-nvi @nv-kmcgill53 I had to rebase the PR. If you have any questions, let me know. Thank you both for reviewing!

jbkyang-nvi (approved):

PR #200, opened 2023-07-06
OpenVINO 2023.0.0

kthui:
server: https://github.com/triton-inference-server/server/pull/6031
openvino_backend: https://github.com/triton-inference-server/openvino_backend/pull/57

Some notes on the upgrade:
- The `tbb.so` library no longer needs to be linked, so it is removed from the build.
- There are quite a few extra libraries built but not required for CPU acceleration (i.e. PaddlePaddle), so only the required libraries are included to save final image size.

### DIFF
--- a/tools/gen_ort_dockerfile.py
+++ b/tools/gen_ort_dockerfile.py
@@ -125,13 +125,9 @@ RUN _CUDNN_VERSION=$(echo $CUDNN_VERSION | cut -d. -f1-2) && \
 # Install OpenVINO
 ARG ONNXRUNTIME_OPENVINO_VERSION
 ENV INTEL_OPENVINO_DIR /opt/intel/openvino_${ONNXRUNTIME_OPENVINO_VERSION}
-ENV InferenceEngine_DIR=$INTEL_OPENVINO_DIR/runtime/cmake
-ENV ngraph_DIR=$INTEL_OPENVINO_DIR/runtime/cmake
-ENV OpenVINO_DIR=$INTEL_OPENVINO_DIR/runtime/cmake
-ENV LD_LIBRARY_PATH $INTEL_OPENVINO_DIR/runtime/lib/intel64:$LD_LIBRARY_PATH
-ENV PKG_CONFIG_PATH=$INTEL_OPENVINO_DIR/runtime/lib/intel64/pkgconfig
-ENV PYTHONPATH $INTEL_OPENVINO_DIR/python/python3.10:$INTEL_OPENVINO_DIR/python/python3:$PYTHONPATH
 
+# Step 1: Download and install core components
+# Ref: https://docs.openvino.ai/2023.0/openvino_docs_install_guides_installing_openvino_from_archive_linux.html#step-1-download-and-install-the-openvino-core-components
 RUN curl -L https://storage.openvinotoolkit.org/repositories/openvino/packages/2023.0/linux/l_openvino_toolkit_ubuntu22_2023.0.0.10926.b4452d56304_x86_64.tgz --output openvino_${ONNXRUNTIME_OPENVINO_VERSION}.tgz && \
     tar -xf openvino_${ONNXRUNTIME_OPENVINO_VERSION}.tgz && \
     mkdir -p ${INTEL_OPENVINO_DIR} && \
@@ -140,6 +136,15 @@ RUN curl -L https://storage.openvinotoolkit.org/repositories/openvino/packages/2
     (cd ${INTEL_OPENVINO_DIR}/install_dependencies && \
         ./install_openvino_dependencies.sh -y) && \
     ln -s ${INTEL_OPENVINO_DIR} ${INTEL_OPENVINO_DIR}/../openvino_`echo ${ONNXRUNTIME_OPENVINO_VERSION} | awk '{print substr($0,0,4)}'`
+
+# Step 2: Configure the environment
+# Ref: https://docs.openvino.ai/2023.0/openvino_docs_install_guides_installing_openvino_from_archive_linux.html#step-2-configure-the-environment
+ENV InferenceEngine_DIR=$INTEL_OPENVINO_DIR/runtime/cmake
+ENV ngraph_DIR=$INTEL_OPENVINO_DIR/runtime/cmake
+ENV OpenVINO_DIR=$INTEL_OPENVINO_DIR/runtime/cmake
+ENV LD_LIBRARY_PATH $INTEL_OPENVINO_DIR/runtime/lib/intel64:$LD_LIBRARY_PATH
+ENV PKG_CONFIG_PATH=$INTEL_OPENVINO_DIR/runtime/lib/intel64/pkgconfig
+ENV PYTHONPATH $INTEL_OPENVINO_DIR/python/python3.10:$INTEL_OPENVINO_DIR/python/python3:$PYTHONPATH
 '''
 
 
--- a/tools/gen_ort_dockerfile.py
+++ b/tools/gen_ort_dockerfile.py
@@ -310,19 +310,17 @@ RUN cp /workspace/build/${ONNXRUNTIME_BUILD_CONFIG}/libonnxruntime_providers_ope
        /opt/onnxruntime/lib && \
     cp /usr/lib/x86_64-linux-gnu/libpugixml.so.1 /opt/onnxruntime/lib
 
-RUN OV_SHORT_VERSION=`echo ${ONNXRUNTIME_OPENVINO_VERSION} | awk '{print substr($0,3,2)}'` && \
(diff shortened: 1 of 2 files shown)


tools/gen_ort_dockerfile.py
```diff
@@ -125,31 +125,21 @@ def dockerfile_for_linux(output_file):
...
-    cd ${INTEL_OPENVINO_DIR}/install_dependencies && os=ubuntu20.04 ./install_openvino_dependencies.sh -y
-
-ARG INTEL_COMPUTE_RUNTIME_URL=https://github.com/intel/compute-runtime/releases/download/19.41.14441
-RUN wget ${INTEL_COMPUTE_RUNTIME_URL}/intel-gmmlib_19.3.2_amd64.deb && \
-    wget ${INTEL_COMPUTE_RUNTIME_URL}/intel-igc-core_1.0.2597_amd64.deb && \
-    wget ${INTEL_COMPUTE_RUNTIME_URL}/intel-igc-opencl_1.0.2597_amd64.deb && \
-    wget ${INTEL_COMPUTE_RUNTIME_URL}/intel-opencl_19.41.14441_amd64.deb && \
-    wget ${INTEL_COMPUTE_RUNTIME_URL}/intel-ocloc_19.41.14441_amd64.deb && \
-    dpkg -i *.deb && rm -rf *.deb
+ENV OpenVINO_DIR=$INTEL_OPENVINO_DIR/runtime/cmake
```
nv-kmcgill53, line 130: What is using this environment variable, is it for openvino or onnxruntime? Is it needed?

tools/gen_ort_dockerfile.py
```diff
@@ -125,31 +125,21 @@ def dockerfile_for_linux(output_file):
...
+ENV PKG_CONFIG_PATH=$INTEL_OPENVINO_DIR/runtime/lib/intel64/pkgconfig
+ENV PYTHONPATH $INTEL_OPENVINO_DIR/python/python3.10:$INTEL_OPENVINO_DIR/python/python3:$PYTHONPATH
+
+RUN curl -L https://storage.openvinotoolkit.org/repositories/openvino/packages/2023.0/linux/l_openvino_toolkit_ubuntu22_2023.0.0.10926.b4452d56304_x86_64.tgz --output openvino_${ONNXRUNTIME_OPENVINO_VERSION}.tgz && \
+    tar -xf openvino_${ONNXRUNTIME_OPENVINO_VERSION}.tgz && \
+    mkdir -p ${INTEL_OPENVINO_DIR} && \
+    mv l_openvino_toolkit_ubuntu22_2023.0.0.10926.b4452d56304_x86_64/* ${INTEL_OPENVINO_DIR} && \
+    rm openvino_${ONNXRUNTIME_OPENVINO_VERSION}.tgz && \
+    (cd ${INTEL_OPENVINO_DIR}/install_dependencies && \
+        os=ubuntu22.04 ./install_openvino_dependencies.sh -y) && \
```
nnshah1, line 141: the os override was added for initial ubunu22.04 support - would recommend to remove it and ensure openvino os deetection is working correctly (to avoid hiding any issues in the future with os changes)

tools/gen_ort_dockerfile.py
```diff
@@ -303,23 +293,31 @@ def dockerfile_for_linux(output_file):
...
+
+RUN OV_SHORT_VERSION=`echo ${ONNXRUNTIME_OPENVINO_VERSION} | awk '{print substr($0,3,2)}'` && \
+    OV_SHORT_VERSION=${OV_SHORT_VERSION}`echo ${ONNXRUNTIME_OPENVINO_VERSION} | awk '{print substr($0,6,1)}'` && \
+    OV_SHORT_VERSION=${OV_SHORT_VERSION}`echo ${ONNXRUNTIME_OPENVINO_VERSION} | awk '{print substr($0,8,1)}'` && \
     (cd /opt/onnxruntime/lib && \
-     chmod a-x * && \
-     ln -sf libtbb.so.2 libtbb.so)
+        chmod a-x * && \
+        ln -s libopenvino.so.${ONNXRUNTIME_OPENVINO_VERSION} libopenvino.so.${OV_SHORT_VERSION} && \
+        ln -s libopenvino.so.${OV_SHORT_VERSION} libopenvino.so && \
```
nv-kmcgill53, line 314: Why is this indirection needed? Why can't we symbolically link the `.so` libs directly?

tools/gen_ort_dockerfile.py
```diff
@@ -303,23 +293,31 @@ def dockerfile_for_linux(output_file):
...
-    cp ${INTEL_OPENVINO_DIR}/runtime/lib/intel64/plugins.xml \
-       /opt/onnxruntime/lib && \
-    cp ${INTEL_OPENVINO_DIR}/runtime/3rdparty/tbb/lib/libtbb.so.2 \
+    cp ${INTEL_OPENVINO_DIR}/runtime/lib/intel64/libopenvino_onnx_frontend.so.${ONNXRUNTIME_OPENVINO_VERSION} \
        /opt/onnxruntime/lib && \
+    cp /usr/lib/x86_64-linux-gnu/libpugixml.so.1 /opt/onnxruntime/lib
+
+RUN OV_SHORT_VERSION=`echo ${ONNXRUNTIME_OPENVINO_VERSION} | awk '{print substr($0,3,2)}'` && \
+    OV_SHORT_VERSION=${OV_SHORT_VERSION}`echo ${ONNXRUNTIME_OPENVINO_VERSION} | awk '{print substr($0,6,1)}'` && \
+    OV_SHORT_VERSION=${OV_SHORT_VERSION}`echo ${ONNXRUNTIME_OPENVINO_VERSION} | awk '{print substr($0,8,1)}'` && \
```
nv-kmcgill53, line 310: As commented below, I am not sure this is needed but you can compress this into one line if I understand correctly:
```
# 2023.0.0 --> 2300
OV_SHORT_VERSION=`echo ${ONNXRUNTIME_OPENVNO_VERSION} | awk '{ split($0,a,"."); print substr(a[1],3) a[2] a[3] }'`
```

This also has the benefit of capturing multi-digit minor and patch versions

nv-kmcgill53 (changes requested):

tanmayv25 (dismissed):

tools/gen_ort_dockerfile.py
kthui: Yes. It is removed: [Remove OS override](https://github.com/triton-inference-server/onnxruntime_backend/pull/200/commits/3980f049bdb7634206c2e309e396b26bf8fa971b)

tools/gen_ort_dockerfile.py
kthui: They are from the OV installation steps and required by the ORT build script to locate the OV build. I updated the script to better reflect where those environment variables are from: [Clarify OV install steps](https://github.com/triton-inference-server/onnxruntime_backend/pull/200/commits/8d56eb9a08615f03de391fdd268829508cda69de)

tools/gen_ort_dockerfile.py
kthui: The indirection is how the OV library are linked by default, but I do not think there are any issues linking them directly: [Improve OV short version generation and link so files directly](https://github.com/triton-inference-server/onnxruntime_backend/pull/200/commits/4d7e46a9d3a4efb7d3a716a47bbdf69b3ed961ef)

tools/gen_ort_dockerfile.py
kthui: This looks better, updated: [Improve OV short version generation and link so files directly](https://github.com/triton-inference-server/onnxruntime_backend/pull/200/commits/4d7e46a9d3a4efb7d3a716a47bbdf69b3ed961ef)

tools/gen_ort_dockerfile.py
nv-kmcgill53: I think I worded my question poorly but your commit shows the intention better :D. My question should have been why do we need `<lib>.so.${OV_SHORT_VERSION}` in the first place, but it seems as though we need all three copies of the file for the compilation. Do I understand this correctly?

nv-kmcgill53 (approved):
Just want an answer to my last question [here](https://github.com/triton-inference-server/onnxruntime_backend/pull/200#discussion_r1255035321). LGTM.

tools/gen_ort_dockerfile.py
kthui: Yes. We need all three copies to compile.

kthui: This is how they are linked originally in OV:
```
...
lrwxrwxrwx 1 root root         19 Jul  6 23:44 libopenvino.so -> libopenvino.so.2300
-rw-r--r-- 1 root root  632499720 Jul  6 23:40 libopenvino.so.2023.0.0
lrwxrwxrwx 1 root root         23 Jul  6 23:44 libopenvino.so.2300 -> libopenvino.so.2023.0.0
...
```

tanmayv25 (approved):

PR #202, opened 2023-07-06
Fix the memory type of output tensor when warm-up is enabled

krishung5:
The issue was that the runtime inference would always return a CPU tensor when warm-up is enabled. This occurred because we recorded the output memory type using the first request. Since the warm-up is the first request to the model and always allocates CPU memory for the output, it incorrectly recorded the memory type for the output.
This PR fixes the logic so that the warm-up inference does not affect the actual memory type of the output tensor.

Testing: https://github.com/triton-inference-server/server/pull/6033


Tabrizian (approved):

GuanLuo (approved):

krishung5:
Rebased.

PR #213, opened 2023-09-20
 Add scalar support in ORT backend

Tabrizian:
testing: https://github.com/triton-inference-server/server/pull/6343

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -885,7 +885,9 @@ ModelState::AutoCompleteIO(const char* key, const OnnxTensorInfoMap& io_infos)
       triton::common::TritonJson::Value reshape_dims(
           ModelConfig(), triton::common::TritonJson::ValueType::ARRAY);
       RETURN_IF_ERROR(reshape.Add("shape", std::move(reshape_dims)));
-      RETURN_IF_ERROR(io.Add("reshape", std::move(reshape)));
+      if (MaxBatchSize() > 0) {
+        RETURN_IF_ERROR(io.Add("reshape", std::move(reshape)));
+      }
     }
     RETURN_IF_ERROR(io.Add("dims", std::move(dims)));
     RETURN_IF_ERROR(ios.Append(std::move(io)));
@@ -997,6 +999,12 @@ class ModelInstanceState : public BackendModelInstance {
       output_device_info_;
   // map of output name -> tensor info
   OnnxTensorInfoMap output_tensor_infos_;
+
+  // map of input name -> tensor info
+  OnnxTensorInfoMap input_tensor_infos_;
+
+  // A map from scalar output tensors to the dimension specified in model config
+  std::unordered_map<std::string, std::vector<int64_t>> scalar_outputs_;
 
   // Onnx Runtime variables that will be reset and used for every run
   // on this instance.
@@ -1313,9 +1321,8 @@ ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
 {
   std::set<std::string> input_tensor_names;
   RETURN_IF_ERROR(InputNames(session_, input_tensor_names));
-
-  OnnxTensorInfoMap input_tensor_infos;
-  RETURN_IF_ERROR(InputInfos(session_, default_allocator_, input_tensor_infos));
+  RETURN_IF_ERROR(
+      InputInfos(session_, default_allocator_, input_tensor_infos_));
 
   std::set<std::string> overridable_initializer_tensor_names;
   RETURN_IF_ERROR(OverridableInitializerNames(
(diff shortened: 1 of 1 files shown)


src/onnxruntime.cc
```diff
@@ -1900,13 +1951,34 @@ ModelInstanceState::SetInputTensors(
...
+      }
+
       // Create ORT Tensor
-      RETURN_IF_ORT_ERROR(ort_api->CreateTensorWithDataAsOrtValue(
-          memory_type == TRITONSERVER_MEMORY_GPU ? cuda_allocator_info_
-                                                 : cpu_allocator_info_,
-          (void*)input_buffer, batchn_byte_size, batchn_shape.data(),
-          batchn_shape.size(), ConvertToOnnxDataType(input_datatype),
-          &input_tensors_.back()));
+      if (iti->second.dims_.size() == 0) {
```
rmccorm4, line 1966: Why are scalar dims empty/size-zero? Shouldn't it be `[1]` or any `[1, ..., 1]` where the "volume" is 1?

Tabrizian: From Triton's perspective, what you mentioned is correct. From model's perspective (which is the condition here), it has zero dimensions.

src/onnxruntime.cc
```diff
@@ -1419,9 +1428,30 @@ ModelInstanceState::ValidateInputs(const size_t expected_input_cnt)
                 .c_str());
       }
     } else {
-      RETURN_IF_ERROR(CompareDimsSupported(
-          model_state_->Name(), io_name, iit->second.dims_, dims,
-          model_state_->MaxBatchSize(), false /* compare_exact */));
+      if (model_state_->MaxBatchSize() != 0 || iit->second.dims_.size() > 0) {
+        RETURN_IF_ERROR(CompareDimsSupported(
+            model_state_->Name(), io_name, iit->second.dims_, dims,
+            model_state_->MaxBatchSize(), false /* compare_exact */));
```
rmccorm4, line 1434: Can't tell from a glance if this is supported or not, but would it be possible to "scatter" a scalar input to other multi-dimension inputs? 

From client/request side of things, I think Triton core expects that all inputs would have a batch dimension - but I wonder if we can generalize scalar inputs from client side to acceptable "tensor" shapes based on other inputs?

src/onnxruntime.cc
```diff
@@ -1482,9 +1512,30 @@ ModelInstanceState::ValidateOutputs()
 
     // The batch output shape doesn't necessarily match the model
     if (model_state_->FindBatchOutput(io_name) == nullptr) {
-      RETURN_IF_ERROR(CompareDimsSupported(
-          model_state_->Name(), io_name, iit->second.dims_, dims,
-          model_state_->MaxBatchSize(), true /* compare_exact */));
+      // if max_batch_size == 0 and is a scalar tensor all the
+      // dimensions specified must be equal to 1
+      if (model_state_->MaxBatchSize() > 0 || iit->second.dims_.size() > 0) {
```
rmccorm4, line 1517: Summarizing offline discussion, I'd rather see `MaxBatchSize()` removed from this condition of "is_scalar" if not necesary. Maybe something like this:
```cpp
const bool is_scalar = iit->second.dims_.size() == 0;
if (is_scalar) {
  if std::any_of(dims.begin(), dims.end(), [](int dim) { dim != 1}) {
    // ERROR
  }
  scalar_outputs_[io_name] = dims;
} else {
  RETURN_IF_ERROR(CompareDimsSupported(...));
}
```

Tabrizian: fixed.

src/onnxruntime.cc
Tabrizian: discussed offline, an input that only has a batch dimension is already supported.

src/onnxruntime.cc
```diff
@@ -885,7 +885,9 @@ ModelState::AutoCompleteIO(const char* key, const OnnxTensorInfoMap& io_infos)
       triton::common::TritonJson::Value reshape_dims(
           ModelConfig(), triton::common::TritonJson::ValueType::ARRAY);
       RETURN_IF_ERROR(reshape.Add("shape", std::move(reshape_dims)));
-      RETURN_IF_ERROR(io.Add("reshape", std::move(reshape)));
+      if (MaxBatchSize() > 0) {
```
rmccorm4, line 888: Can you add a comment about why reshape causes issues with non-batching case for future reference?

src/onnxruntime.cc
```diff
@@ -2283,6 +2352,22 @@ ModelInstanceState::ReadOutputTensors(
           batchn_shape, dtype, output_tensor, &output_buffer, string_buffers,
           offsets));
 
+      // If the number of dimensions is equal to zero, it means that it is a
+      // scalar and it would use the dimensions specified in the mdel
```
rmccorm4, line 2356:
```suggestion
      // scalar and it would use the dimensions specified in the model
```

src/onnxruntime.cc
rmccorm4: Actually, is this breaking any functionality for non-batching models that specify a "reshape" in their model config? Such as our densenet example: https://github.com/triton-inference-server/server/blob/main/docs/examples/model_repository/densenet_onnx/config.pbtxt

Tabrizian: added a comment.

I don't think it would break it since they all do have some dimensions.

rmccorm4 (approved):

PR #218, opened 2023-11-17
Add ORT patch build

Tabrizian:


tanmayv25 (approved):
Thanks for being pro-active!

PR #221, opened 2023-12-02
Add ORT patch build (#218)

mc-nv:
This PR is redundant, but if someone decide that it need to be approved I will bring it tot the `main`


Tabrizian (commented):
This is not required anymore so let's not merge it to main for now.

PR #248, opened 2024-03-20
Add support for sharing an ORT session

quic-suppugun:
For every instance in a model instance group a new ORT session is created. This code adds support to share a session per instance group.
This support can be enabled by defining
'share_session_between_instances' to true
in triton model config "parameters". Example:
parameters [
.....
  {
    key: "share_session_between_instances"
    value: {string_value: "true"}
  }
]

This is a global parameter and cannot be defined per instance group. The user should determine if the parameter makes sense for their setup.

When log-info option of tritonserver is set to "1", the logs will indicate that a session is mapped for the instance group during the first initialized instance and reused for other instances.
Example:
 TRITONBACKEND_ModelInstanceInitialize: <model>_0_1 (CPU device 0)
 TRITONBACKEND_ModelInstanceInitialize: <model>_0_0 (CPU device 0)
 Could not find a session corresponding to instance group: <model>_0
 Created session for instance: <model>_0_1
 Mapped session for instance group: <model>_0
 Reusing session for instance: <model>_0_0

Change-Id: I6dc509b9c2451e3dd14d45f6f150b37f50b5db89

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -25,7 +25,6 @@
 // OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 #include <stdint.h>
-
 #include <mutex>
 #include <vector>
 
@@ -107,10 +106,10 @@ class ModelState : public BackendModel {
   // onnx file, return in 'session' and 'allocator' the ORT session
   // and allocator.
   TRITONSERVER_Error* LoadModel(
-      const std::string& artifact_name,
+      const std::string& artifact_name, const std::string& instance_name,
       const TRITONSERVER_InstanceGroupKind instance_group_kind,
       const int32_t instance_group_device_id, std::string* model_path,
-      OrtSession** session, OrtAllocator** default_allocator,
+      std::shared_ptr<OrtSession>& session, OrtAllocator** default_allocator,
       cudaStream_t stream);
 
   const std::map<std::string, std::pair<int64_t, int64_t>>& ModelOutputs()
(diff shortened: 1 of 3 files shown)


Jackiexiao:
I have compiled two images based on this PR for easy use. They are:

- docker pull jackiexiao/tritonserver:24.03-py3-onnx-share-session 
- docker pull jackiexiao/tritonserver:24.03-onnx-py-cpu-onnx-share-session

The first image only replaces the ONNX backend while keeping everything else unchanged. The second image provides a smaller CPU version.

adityagoel4512:
Hey! I was going to work on resolving the same issue with session sharing and noticed that this PR already exists, so thanks. Is there a reluctance to do this that I'm missing?

Jacker-bro:
This is great, but it needs to be set with the enable_mem_pattern parameter to solve the performance issue.

glebashnik:
This is really useful and it would be great to merge this.
Since this PR branch is stale I rebased this work on current `main` and cleaned it up a bit with some minor improvements in the new PR:
https://github.com/triton-inference-server/onnxruntime_backend/pull/343                                                                                                                         

It includes:
- The original session-sharing implementation
- Locking improvement for shared session initialization using RAII.
- Improved instance group name parsing without regex. This handles multi-digit instance group indexes and model
names with regex special characters.
- Documented the new parameter.
- Added an integration test that verifies session reuse and runs inference.

I also ran local CPU/GPU benchmarks with Triton Server 2.68.0 and a ModernBERT embedder to measure memory use and throughput. Tested with enable_mem_pattern, do_copy_in_default_stream, and use_ep_level_unified_stream; these did not meaningfully affect performance for shared sessions.

PR #257, opened 2024-06-10
build: Add WAR for CUDA 12.5 build issue

rmccorm4:
Adding ORT patch from here: https://github.com/microsoft/onnxruntime/pull/20770

With build patch we used in a previous release here: https://github.com/triton-inference-server/onnxruntime_backend/commit/05d098d89f0455588fb85f0a22d97dd8ecc6cac9

In case the ORT release containing this fix does not match our timelines for getting r24.06 out.

### DIFF
--- a/tools/gen_ort_dockerfile.py
+++ b/tools/gen_ort_dockerfile.py
@@ -205,8 +205,11 @@ ARG ONNXRUNTIME_VERSION
 ARG ONNXRUNTIME_REPO
 ARG ONNXRUNTIME_BUILD_CONFIG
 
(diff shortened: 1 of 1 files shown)


rmccorm4:
@mc-nv Let me know if this should be targeting `main` + cherry-pick to `r24.06`, or targeting `r24.06` directly if the changes are accepted.

tanmayv25 (approved):

tanmayv25:
Targeting r24.06 will be good enough. We can bring the fix to main once we start building main branches with cuda 12.5

PR #264, opened 2024-08-15
Hardcoding libonnxruntime.so.1 library name reference value

mc-nv:
This is a quick fix to tritonserver container backend. which fix non-found `libonnxruntime.so.1`
```
$ docker run --rm -it  tritonserver:24.08-py3-base bash -c "ldd backends/onnxruntime/libtriton_onnxruntime.so" | grep found
        libtritonserver.so => not found
        libonnxruntime.so.1 => not found
```

Related issue: https://github.com/microsoft/onnxruntime/issues/21763

### DIFF
--- a/tools/gen_ort_dockerfile.py
+++ b/tools/gen_ort_dockerfile.py
@@ -251,7 +251,7 @@ RUN git clone -b rel-${ONNXRUNTIME_VERSION} --recursive ${ONNXRUNTIME_REPO} onnx
 
     df += """
 WORKDIR /workspace/onnxruntime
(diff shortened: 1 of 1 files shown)


pvijayakrish:
Do we have a RCA? Is this a temporary fix to unblock QA for 24.08 testing?

mc-nv:
> Do we have a RCA? Is this a temporary fix to unblock QA for 24.08 testing?

We have to accept change and rebuild the image to provide image to QA

pvijayakrish:
Approving the temporary fix to unblock QA for 24.08 testing. Please create a follow up ticket, if required, to investigate and adopt a permanent solution.

pvijayakrish (approved):

pvijayakrish (approved):

PR #273, opened 2024-09-26
Update ORT repo post release 24.09 

pvijayakrish:
Includes the following changes: 

1. Increase Docker limits on Windows 
2. hardcoding missing sym link library reference for ONNX


mc-nv (approved):

CMakeLists.txt
```diff
@@ -157,7 +157,7 @@ else()
   endif()
 
   set(TRITON_ONNXRUNTIME_DOCKER_IMAGE "tritonserver_onnxruntime")
-  set(TRITON_ONNXRUNTIME_DOCKER_MEMORY "8g")
+  set(TRITON_ONNXRUNTIME_DOCKER_MEMORY "$<IF:$<BOOL:WIN32>,32g,8g>")
```
nv-kmcgill53, line 160: Does this still work if a windows machine doesn't have 32g? Does the compilation just take all the available memory instead?

nv-kmcgill53 (approved):

CMakeLists.txt
mc-nv: It's a docker option if it can't allocate given amount it should use default limits. 
Because it's a cached argument user can always change it as per their needs.

PR #276, opened 2024-10-18
Use pre-built ONNX Runtime binaries.

mc-nv:
server: https://github.com/triton-inference-server/server/pull/7717


nv-kmcgill53 (approved):

cmake/download_onnxruntime.cmake
```diff
@@ -0,0 +1,25 @@
...
+
+  if(NOT DOWNLOAD_RESULT EQUAL 0)
+    message(NOTICE "Failed to download: ${TRITON_ONNXRUNTIME_PACKAGE_URL}")
+  else()
+    message(NOTICE "Download successful: ${DOWNLOAD_PATH}" )
+
+    file(ARCHIVE_EXTRACT INPUT ${DOWNLOAD_PATH} DESTINATION ${EXTRACT_DIR} VERBOSE )
+
+    set(TRITON_ONNXRUNTIME_INCLUDE_PATHS ${EXTRACT_DIR}/include)
+    set(TRITON_ONNXRUNTIME_LIB_PATHS ${EXTRACT_DIR}/lib)
```
fpetrini15, line 21: Is this going to break the RHEL build?

mc-nv: No unless you start feed the prebuilt asset for RHEL.

fpetrini15: We may end up using this for our Linux builds in the future, but I suppose we can enhance it when needed.

mc-nv: Is it mean that we can resolve this thread?

fpetrini15 (approved):

cmake/download_onnxruntime.cmake
```diff
@@ -0,0 +1,25 @@
+if(DEFINED TRITON_ONNXRUNTIME_PACKAGE_URL)
+
+  set(DOWNLOAD_PATH "${CMAKE_BINARY_DIR}/_deps/downloads/onnxruntime.zip")
```
GuanLuo, line 3: How do we get the original `onnxruntime.zip`? Is that a separate (optional) build step in our pipeline?

cmake/download_onnxruntime.cmake
```diff
@@ -0,0 +1,25 @@
...
+
+  # file(DOWNLOAD ... STATUS DOWNLOAD_STATUS) returns a list with 2 elements
+  list(GET DOWNLOAD_STATUS 0 DOWNLOAD_RESULT)
+
+  if(NOT DOWNLOAD_RESULT EQUAL 0)
+    message(NOTICE "Failed to download: ${TRITON_ONNXRUNTIME_PACKAGE_URL}")
+  else()
+    message(NOTICE "Download successful: ${DOWNLOAD_PATH}" )
+
+    file(ARCHIVE_EXTRACT INPUT ${DOWNLOAD_PATH} DESTINATION ${EXTRACT_DIR} VERBOSE )
```
GuanLuo, line 18: Do we care about the ORT version? If so, perhaps we should package the version in a file in the compressed artifact and check if the version matches?

cmake/download_onnxruntime.cmake
mc-nv: It's manual for now.

cmake/download_onnxruntime.cmake
mc-nv: We do care about version. 
Version is compressed. 
I can add validator if it's your ask, please raise a ticket against me.

mc-nv: We can put a hack like this: 
```CMake 
        file(READ "${EXTRACT_DIR}/VERSION_NUMBER" DOWNLOADED_ONNXRUNTIME_VERSION)
        if( NOT ${DOWNLOADED_ONNXRUNTIME_VERSION} VERSION_EQUAL ${TRITON_BUILD_ONNXRUNTIME_VERSION})
          set(TRITON_ONNXRUNTIME_INCLUDE_PATHS ${EXTRACT_DIR}/include)
          set(TRITON_ONNXRUNTIME_LIB_PATHS ${EXTRACT_DIR}/lib)
        endif()
```

GuanLuo: I think this is sufficient for now

mc-nv: I don't believe that I understand you comment in right, does it mean that we can close the thread? 
Or you willing me to extend PR with proposed above code snippet?

GuanLuo: Would be great if you can extend PR with proposed above code snippet

GuanLuo (approved):

PR #283, opened 2024-11-08
Enable support for Ubuntu 24.04

mc-nv:
This PR enables support for Ubuntu 24.04 by adding a new job to the CI workflow.

### DIFF
--- a/tools/gen_ort_dockerfile.py
+++ b/tools/gen_ort_dockerfile.py
@@ -106,7 +106,7 @@ ENV PIP_BREAK_SYSTEM_PACKAGES=1
         df += """
 # The manylinux container defaults to Python 3.7, but some feature installation
 # requires a higher version.
-ARG PYVER=3.10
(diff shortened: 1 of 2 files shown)


tools/gen_ort_dockerfile.py
```diff
@@ -105,7 +106,7 @@ def dockerfile_for_linux(output_file):
         df += """
 # The manylinux container defaults to Python 3.7, but some feature installation
```
nv-anants, line 107: nit: comment still accurate?

nv-anants (approved):

tools/gen_ort_dockerfile.py
mc-nv: have no idea, cc @fpetrini15

fpetrini15: Yes, this comment is still accurate

PR #285, opened 2024-11-18
build: Support RHEL ORT TensorRT Execution Provider

fpetrini15:
**Goal:** Build adjustments to enable the TensorRT execution provider for the RHEL ORT backend.


CMakeLists.txt
```diff
@@ -137,6 +137,15 @@ if(TRITON_ONNXRUNTIME_LIB_PATHS STREQUAL "")
...
 endif()
 
+set(RHEL_BUILD OFF)
+if(LINUX)
+  file(STRINGS "/etc/os-release" DISTRO_ID_LIKE REGEX "ID_LIKE")
+  if(${DISTRO_ID_LIKE} MATCHES "rhel|centos")
+    set(RHEL_BUILD ON)
+  endif(${DISTRO_ID_LIKE} MATCHES "rhel|centos")
+endif(LINUX)
+set(TRITON_CORE_HEADERS_ONLY OFF)
```
nv-kmcgill53, line 147: Why is this needed? I thought this is only needed for building the `core` repository?

CMakeLists.txt
```diff
@@ -324,6 +333,13 @@ endif() # TRITON_ENABLE_ONNXRUNTIME_OPENVINO
 #
 if(TRITON_ONNXRUNTIME_DOCKER_BUILD)
   set(_GEN_FLAGS "")
+  if(${RHEL_BUILD} AND "${TRITON_BUILD_TENSORRT_HOME}" STREQUAL "")
+    if(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "aarch64")
+      set(TRITON_BUILD_TENSORRT_HOME "/usr/local/cuda/targets/sbsa-linux/")
+    else()
+      set(TRITON_BUILD_TENSORRT_HOME "/usr/local/cuda/targets/x86_64-linux/")
+    endif(${CMAKE_SYSTEM_PROCESSOR} STREQUAL "aarch64")
+  endif(${RHEL_BUILD} AND "${TRITON_BUILD_TENSORRT_HOME}" STREQUAL "")
```
nv-kmcgill53, line 342: I believe the current behavior is that compilation fails if the `--tensorrt-home` flag is not set. So is setting a default required here? I think we should have this fail if `gen_ort_dockerfile.py` doesn't have `--tensorrt-home` set. If ware keeping this logic in then please move:
```
if(NOT ${TRITON_BUILD_TENSORRT_HOME} STREQUAL "")
    set(_GEN_FLAGS ${_GEN_FLAGS} "--tensorrt-home=${TRITON_BUILD_TENSORRT_HOME}")
  endif() # TRITON_BUILD_TENSORRT_HOME
```
up to right below this logic so it's clearer what is being set. Thanks.

CMakeLists.txt
fpetrini15: Totally not needed. This was a copy-paste error.

CMakeLists.txt
fpetrini15: That is correct, though it only seems to fail the RHEL builds if this flag is not provided. During development, I struggled with whether this logic should be handled in the CMake directly or if it should be passed in via `build.py`. Do you have a preference?

nv-kmcgill53: This was hard to track down but on [gen_ort_dockerfile.py:575](https://github.com/triton-inference-server/onnxruntime_backend/blob/main/tools/gen_ort_dockerfile.py#L574-L575) we can see that `tensorrt_home` is being set to a default when the parameter is not found. Can we move your default logic out to `gen_ort_dockerfile.py` instead? I think this would be the right place to put this since we can conceivably compile onnxruntime without the enable gpu flags and then we wouldn't want to set a default here (I think).

fpetrini15: Ironically, I had also tried adding this logic exactly where you're pointing to, however, I couldn't find a reliable way to get Python to detect an arm64 machine and didn't want to add a new flag. Let me give it a second look.

nv-kmcgill53: I see. Well, give this another 30 min of work to see if fresh eyes helps out. If not then, just group the cmake logic together for the setting of the `TRITON_BUILD_TENSORRT_HOME` variable.

fpetrini15: From what I can tell, it is technically possible for arm-based systems to return something other than `aarch64`, however, `platform.machine()` is a convention we use in other places and should be fine.

nv-kmcgill53 (approved):

PR #312, opened 2025-07-09
build: TPRD-1622 - Update version to CMake 4.0 or 3.31 ( 3.31.8 ) whi…

mc-nv:
…ch has support for CUDA new layout

### DIFF
--- a/tools/gen_ort_dockerfile.py
+++ b/tools/gen_ort_dockerfile.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# Copyright 2020-2024, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# Copyright 2020-2025, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 #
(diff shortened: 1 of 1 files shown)


tools/gen_ort_dockerfile.py
```diff
@@ -177,18 +177,7 @@ def dockerfile_for_linux(output_file):
         gnupg \
         gnupg1
 
-RUN pip3 install patchelf==0.17.2
-
-# Install dependencies from
```
yinggeh, line 182: We don't need this any more?

mc-nv: We can install it as a python package.

tools/gen_ort_dockerfile.py
```diff
@@ -177,18 +177,7 @@ def dockerfile_for_linux(output_file):
         gnupg \
         gnupg1
 
-RUN pip3 install patchelf==0.17.2
```
yinggeh, line 180: Update copyrights

yinggeh (approved):

PR #354, opened 2026-08-10
improve(Build): cap ort build parallellism with explicit value

jcuquemelle:
#### What does the PR do?
This allows the ORT build to accept a lower parallelism than the heuristics from https://github.com/triton-inference-server/onnxruntime_backend/pull/348

Mainly to propagate the explicit parallelism value from build.py

#### Checklist
- [x] I have read the [Contribution guidelines](#../../CONTRIBUTING.md) and signed the [Contributor License
Agreement](https://github.com/NVIDIA/triton-inference-server/blob/master/Triton-CCLA-v1.pdf)
- [x] PR title reflects the change and is of format `<commit_type>: <Title>`
- [x] Changes are described in the pull request.
- [x] Related issues are referenced.
- [ ] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [x] Added [test plan](#test-plan) and verified test passes.
- [ ] Verified that the PR passes existing CI.
- [x] I ran pre-commit locally (`pre-commit install, pre-commit run --all`)
- [x] Verified copyright is correct on all changed files.
- [ ] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [x] All template sections are filled out.

#### Commit Type:
Check the [conventional commit type](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)
box here and add the label to the github PR.
- [x] build
- [ ] ci
- [ ] docs
- [ ] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Related PRs:
https://github.com/triton-inference-server/server/pull/8926

#### Test plan:
Run the main triton build (build.py) with related options 
Tested on internal CI

#### Caveats:
None

#### Background
See related PR on triton-server

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves /


CMakeLists.txt
```diff
@@ -345,6 +348,17 @@ if(TRITON_ONNXRUNTIME_DOCKER_BUILD)
   if(${TRITON_ENABLE_ONNXRUNTIME_OPENVINO})
     set(_GEN_FLAGS ${_GEN_FLAGS} "--ort-openvino=${TRITON_BUILD_ONNXRUNTIME_OPENVINO_VERSION}")
   endif() # TRITON_ENABLE_ONNXRUNTIME_OPENVINO
+  if(NOT TRITON_BUILD_PARALLEL STREQUAL "")
+    set(_GEN_FLAGS ${_GEN_FLAGS} "--parallel-jobs=${TRITON_BUILD_PARALLEL}")
+  endif() # TRITON_BUILD_PARALLEL
```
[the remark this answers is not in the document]
jcuquemelle: The cuda arch list is forwarded by an env var which is handled at ord dockerfile creation here https://github.com/triton-inference-server/onnxruntime_backend/pull/354/changes#diff-ad5e480d1a7be6ef5d700f428d3f7da2559e36ee630e6a848a959dcdc3753832L445

It is not needed to add it to CMakeLists

CMakeLists.txt
```diff
@@ -345,6 +349,20 @@ if(TRITON_ONNXRUNTIME_DOCKER_BUILD)
...
+  if(NOT TRITON_CUDA_ARCH_LIST STREQUAL "")
+    set(_GEN_FLAGS ${_GEN_FLAGS} "--cuda-arch-list=${TRITON_CUDA_ARCH_LIST}")
```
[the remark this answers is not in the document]
jcuquemelle: TRITON_CUDA_ARCH_LIST was redundant (information passes through CUDA_ARCH_LIST  env var, so these lines have been removed

tools/gen_ort_dockerfile.py
```diff
@@ -462,6 +469,11 @@ def dockerfile_for_linux(output_file):
     # the build stage stays free of shell logic (portable across Debian/RHEL).
     nvcc_threads = 2
     ort_jobs = build_parallelism(nvcc_threads)
+    # Honor explicit --parallel-jobs override (otherwise the TRI-1550 cap above
+    # decides); r26.07 passes `--parallel {ort_jobs}` on the RUN line, so the
+    # value computed here is what the build actually uses.
+    if FLAGS.parallel_jobs is not None:
```
Vinya567, line 475: When `--parallel-jobs` is set, this overwrites `ort_jobs = build_parallelism(...)`. We would like to keep that limit so nested ORT compile does not run the builder out of memory. If the limit is too tight for you, can you describe the setup?

CMakeLists.txt
```diff
@@ -356,7 +370,7 @@ if(TRITON_ONNXRUNTIME_DOCKER_BUILD)
       OUTPUT
         onnxruntime/lib/${ONNXRUNTIME_LIBRARY}
       COMMAND python3 ${CMAKE_CURRENT_SOURCE_DIR}/tools/gen_ort_dockerfile.py --triton-container="${TRITON_BUILD_CONTAINER}" --ort-version="${TRITON_BUILD_ONNXRUNTIME_VERSION}" --trt-version="${TRT_VERSION}" --onnx-tensorrt-tag="${TRITON_ONNX_TENSORRT_REPO_TAG}" ${_GEN_FLAGS} --output=Dockerfile.ort ${ENABLE_GPU_EXTRA_ARGS}
-      COMMAND docker build --memory ${TRITON_ONNXRUNTIME_DOCKER_MEMORY} --cache-from=${TRITON_ONNXRUNTIME_DOCKER_IMAGE} --cache-from=${TRITON_ONNXRUNTIME_DOCKER_IMAGE}_cache0 --cache-from=${TRITON_ONNXRUNTIME_DOCKER_IMAGE}_cache1 -t ${TRITON_ONNXRUNTIME_DOCKER_IMAGE} -f ./Dockerfile.ort ${CMAKE_CURRENT_SOURCE_DIR}
+      COMMAND docker build ${_DOCKER_BUILD_NETWORK_ARGS} --memory ${TRITON_ONNXRUNTIME_DOCKER_MEMORY} --cache-from=${TRITON_ONNXRUNTIME_DOCKER_IMAGE} --cache-from=${TRITON_ONNXRUNTIME_DOCKER_IMAGE}_cache0 --cache-from=${TRITON_ONNXRUNTIME_DOCKER_IMAGE}_cache1 -t ${TRITON_ONNXRUNTIME_DOCKER_IMAGE} -f ./Dockerfile.ort ${CMAKE_CURRENT_SOURCE_DIR}
       COMMAND powershell.exe -noprofile -c "docker rm onnxruntime_backend_ort > $null 2>&1; if ($LASTEXITCODE) { 'error ignored...' }; exit 0"
```
Vinya567, line 374: Optional extra `docker build --network=...` arg, default empty, looks reasonable. What does `--network` cover in your environment?

Vinya567:
Thanks for the PR.

The description mentions a CUDA arch list and `DOCKER_HOST`. Those are not in this diff (`CMakeLists.txt` and `tools/gen_ort_dockerfile.py` only). Could you update the description to match?

tools/gen_ort_dockerfile.py
```diff
@@ -364,6 +364,13 @@ def dockerfile_for_linux(output_file):
             openvino_toolkit_filename, openvino_folder_name
         )
 
+    if FLAGS.no_root_build:
+        df += """
+RUN useradd --create-home --shell /bin/bash tritonbuild && \\
+    chown -R tritonbuild:tritonbuild /workspace
+USER tritonbuild
```
Vinya567, line 371: Default stays root, which is good. Why the non-root `USER` on this intermediate image? We would want to test this path before taking it.

tools/gen_ort_dockerfile.py
mc-nv, line 1: May I know motivation behind changing file that is no more than wrapper for the https://github.com/microsoft/onnxruntime/ build?

CMakeLists.txt
mc-nv, line 1: 1. We can addd flexibility and specific user customization, but suggested approach comes with redundancy at glance. 
2. I don't really want to set docker engine configuration out of CMake explicitly especially considering that we running docker-out-of-docker.

tools/gen_ort_dockerfile.py
jcuquemelle: Originally I started this contribution before the inner ORT build_parallelism() was implemented, so forwarding --parallel-jobs was the way to reduce the nested build pressure. 
Now the new mechanism mitigates the issue, but the author states in their proposal that the 2GB per compilation job could not work in all cases, which is why I kept this possibility to reduce the parallelism of the nested build from the triton server build script. 

I would propose to do a min(FLAGS.parallel_jobs, build_parallelism(...)) to keep this possibility, does that sound good ?

CMakeLists.txt
jcuquemelle: either empty, or --network=host (necessary for non-root builds)

PR #355, opened 2026-08-11
feat: Wire trt_op_types_to_exclude from Triton ort backend to ORT runtime

jcuquemelle:
#### What does the PR do?
Some Onnx models using operators with a cuda implementation but no corresponding TRT implementation can fail to run on TRT EP if the list of unsupported operators is not provided through this flag

example of error: `[E:onnxruntime:, inference_session.cc:2805 operator()] Exception during initialization: /workspace/onnxruntime/onnxruntime/core/providers/tensorrt/tensorrt_execution_provider.cc:2308 SubGraphCollection_t onnxruntime::TensorrtExecutionProvider::GetSupportedList(SubGraphCollection_t, int, int, const onnxruntime::GraphViewer&, bool*) const [ONNXRuntimeError] : 1 : FAIL : Fatal error: this:categoricallabel_encoder_int(-1) is not a registered function/op`

This PR allows to solve the issue by adding this parameter in the TRT configuration block of `config.pbtxt`:
```
optimization {
    execution_accelerators {
      gpu_execution_accelerator: [
        {
          name: "tensorrt"
          .....
          parameters { key: "trt_op_types_to_exclude"  value: "categoricallabel_encoder_int,multicategoricallabel_encoder_int" }
        }
      ]
    }
  }
```

#### Checklist
- [x] I have read the [Contribution guidelines](#../../CONTRIBUTING.md) and signed the [Contributor License
Agreement](https://github.com/NVIDIA/triton-inference-server/blob/master/Triton-CCLA-v1.pdf)
- [x] PR title reflects the change and is of format `<commit_type>: <Title>`
- [x] Changes are described in the pull request.
- [x] Related issues are referenced.
- [ ] Populated [github labels](https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels) field
- [x] Added [test plan](#test-plan) and verified test passes.
- [ ] Verified that the PR passes existing CI.
- [x] I ran pre-commit locally (`pre-commit install, pre-commit run --all`) 
- [x] Verified copyright is correct on all changed files.
- [ ] Added _succinct_ git squash message before merging [ref](https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html).
- [x] All template sections are filled out.
- [ ] Optional: Additional screenshots for behavior/output changes with before/after.

#### Commit Type:
Check the [conventional commit type](https://github.com/angular/angular/blob/22b96b9/CONTRIBUTING.md#type)
box here and add the label to the github PR.
- [] build
- [ ] ci
- [ ] docs
- [x] feat
- [ ] fix
- [ ] perf
- [ ] refactor
- [ ] revert
- [ ] style
- [ ] test

#### Where should the reviewer start?
Only 1 file changed in the PR

#### Test plan:
Rebuild the server, serve an Onnx model that contains a [labelEncoder](https://onnx.ai/onnx/operators/onnx_aionnxml_LabelEncoder.html) operator using the TRT execution provider.
Tested internally

#### Related Issues: (use one of the action keywords Closes / Fixes / Resolves / Relates to)
None

### DIFF
--- a/src/onnxruntime.cc
+++ b/src/onnxruntime.cc
@@ -557,6 +557,10 @@ ModelState::LoadModel(
                     RETURN_IF_ERROR(
                         params.MemberAsString(param_key.c_str(), &value));
                     key = "trt_engine_cache_path";
+                  } else if (param_key == "trt_op_types_to_exclude") {
+                    RETURN_IF_ERROR(
+                        params.MemberAsString(param_key.c_str(), &value));
+                    key = "trt_op_types_to_exclude";
                   } else if (param_key == "trt_engine_cache_prefix") {
                     RETURN_IF_ERROR(
                         params.MemberAsString(param_key.c_str(), &value));
--- a/README.md
+++ b/README.md
@@ -141,6 +141,7 @@ check [ONNX Runtime doc](https://onnxruntime.ai/docs/execution-providers/TensorR
 | trt_dump_ep_context_model         | e.g: "1" or "true", "0" or "false"                  | trt_dump_ep_context_model                        | bool   |
 | trt_ep_context_file_path          |                                                     | trt_ep_context_file_path                         | string |
 | trt_ep_context_embed_mode         | e.g: "1"                                            | trt_ep_context_embed_mode                        | int    |
+| trt_op_types_to_exclude           | e.g. "op_type1,op_type_2"                           | trt_op_types_to_exclude                          | string |
 
 The section of model config file specifying these parameters will look like:
 


Vinya567:
Reviewed. The change looks correct: `trt_op_types_to_exclude` was missing from the TensorRT accelerator allowlist, so it could not be forwarded to ORT.

We ran internal CI with this branch on current `main`. Existing ORT L0s passed, and a config that sets `trt_op_types_to_exclude` now loads (it is no longer rejected as `unknown parameter`). Unknown keys such as `segment_size` still error, as expected.

Two follow-ups, neither is a merge blocker for this wiring:

1. README: please add `trt_op_types_to_exclude` to the TensorRT parameter mapping table (comma-separated op types, same as ORT).
2. Optional: a tiny ONNX + `config.pbtxt` that fails without the flag and loads with it (your `categoricallabel_encoder_int` case). We did not reproduce that model here.

CLA looks good (Julien).

src/onnxruntime.cc
```diff
@@ -557,6 +557,10 @@ ModelState::LoadModel(
                     RETURN_IF_ERROR(
                         params.MemberAsString(param_key.c_str(), &value));
                     key = "trt_engine_cache_path";
+                  } else if (param_key == "trt_op_types_to_exclude") {
+                    RETURN_IF_ERROR(
+                        params.MemberAsString(param_key.c_str(), &value));
+                    key = "trt_op_types_to_exclude";
```
mattwittwer, line 563: Can we also add a line for this to: https://github.com/triton-inference-server/onnxruntime_backend/blob/main/README.md#parameter-mapping-between-onnx-runtime-and-triton-onnxruntime-backend

jcuquemelle: Done

